PR #11, opened 2016-06-21
Trac 5987 fix dynamic linking

Mistuke:
W.I.P


Mistuke:
Sorry, wrong base again.

PR #15, opened 2016-12-02
Test Trac #12919

int-index:
I've encountered a bug in GHC 8.0 and reported it as https://ghc.haskell.org/trac/ghc/ticket/12919

However, it's already fixed in HEAD. This is a test to avoid a regression.


int-index:
https://phabricator.haskell.org/D2788

PR #17, opened 2017-01-31
Fix binary instance for SrcStrictness

alexbiehl:
(hopefully this is a small enough PR)


mpickering:
This looks quite a subtle bug. Is this information used in downstream modules? Does it have an effect on performance somehow?

However,

1. It is not a documentation change.
2. There it skips harbourmaster.
3. You have submitted many patches before using Phab!

alexbiehl:
@mpickering Alright, You can find the Phab revision [here](https://phabricator.haskell.org/D3050).

PR #18, opened 2017-02-01
Fix documentation NOTE about can_fail

int-index:
In the note it is explained why it's fine to discard a can_fail primop, but the table erroneously says it's not fine.


bgamari:
Thanks @int-index. Merged.

PR #20, opened 2017-02-07
Fix docs typo

DanielG:


nomeata:
Thanks, pushed as 078beea. A just slightly more descriptive commit message would be appreciated.

PR #21, opened 2017-02-07
Add Wredundant-constraints to Wall exceptions

siddhanathan:
In light of https://ghc.haskell.org/trac/ghc/ticket/10635


mpickering:
The last commit in that thread says that it was removed from `-Wall`.

siddhanathan:
> The last commit in that thread says that it was removed from -Wall.

Yes, this is a documentation change only.

mpickering:
I see now, this is a list of exceptions which are NOT enabled by `-Wall`.

int-index:
The name of the PR is misleading, you're adding `-Wredundant-constraints` into the list of warnings *not* enabled by `-Wall`.

But the patch appears correct.

nomeata:
Fixed the commit message, pushed as 5ce39f6.

PR #22, opened 2017-02-08
Document the order for sort and sortOn.

alexanderkjeldaas:


alexanderkjeldaas:
Done.  IMO, if the PHP documentation would have documented the property, it should be documented in Haskell.

PR #24, opened 2017-03-03
Fix a tiny typo

iblech:
This pull request fixes a tiny typo.

By the way, I'm wondering what the security implications of the random number generator used for creating temporary file names being entirely predictable (it's just using the PID and a sequential number) are. Probably none on local filesystems thanks to O_EXCL and O_CREAT.

### DIFF
--- a/libraries/base/System/IO.hs
+++ b/libraries/base/System/IO.hs
@@ -426,7 +426,7 @@ fixIO k = do
 -- | The function creates a temporary file in ReadWrite mode.
 -- The created file isn\'t deleted automatically, so you need to delete it manually.
 --
--- The file is creates with permissions such that only the current
+-- The file is created with permissions such that only the current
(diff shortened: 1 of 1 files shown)


bgamari:
Thanks!

bgamari:
> By the way, I'm wondering what the security implications of the random number generator used for creating temporary file names being entirely predictable (it's just using the PID and a sequential number) are. Probably none on local filesystems thanks to O_EXCL and O_CREAT.

Indeed this was the conclusion that was arrived at previously. See [#10940](https://ghc.haskell.org/trac/ghc/ticket/10940).

iblech:
Thanks for the quick merge, the quick reply, and the Trac link! I guess an attacker can only perform a denial of service attack, then. By creating lots of empty files. This is in principle not different to mkstemp, only that there the attacker has to create much more files.

PR #28, opened 2017-03-17
Show valid substitutions for typed holes

Tritlo:
This is currently a work in progress, but I'd like some feedback as I go along (I am totally new to hacking on GHC). The idea is to implement something similar to [type-directed search in purescript](https://github.com/purescript/purescript/pull/2352), i.e. when you leave a hole unfilled, GHC helpfully suggests things in scope that can fit that hole. 

Currently, it finds everything in the local scope and in imported modules that have the same type, and suggests that as a valid substitution. I'd like to extend that to also substitute with any type that can be unified with the given type, so that e.g. for

```
ps :: String -> IO ()
ps = putStrLn

ps2 :: a -> IO ()
ps2 _ = putStrLn "hello, world"

main :: IO ()
main = _ "hello, world"
```

The results would be something like

```
    • Found hole: _ :: [Char] -> IO ()
    • In the expression: _
      In a stmt of a 'do' block: _ "hello, world"
      In the expression:
        do psa "hello, world"
           ps3 "hey"
           _ "hello, world"
    • Relevant bindings include
        main :: IO () (bound at test.hs:13:1)
        ps :: String -> IO () (bound at test.hs:7:1)
        ps2 :: forall a. a  -> IO () (bound at test.hs:10:1)
      Valid substitutions include
        ps :: String -> IO () (defined at test.hs:7:1)
        ps2 :: forall a. a  -> IO () (defined at test.hs:10:1)
        putStrLn :: String
                    -> IO () (imported from ‘Prelude’ at test.hs:1:1-14
                              (and originally defined in ‘System.IO’))
        putStr :: String
                  -> IO () (imported from ‘Prelude’ at test.hs:1:1-14
                            (and originally defined in ‘System.IO’))
```
I've submitted this to Phabricator now, https://phabricator.haskell.org/D3361.

What I have gathered (some experiments can be seen on Tritlo@f4bd18a5da9dfbd7a48b652893172339720a2179), I would need to use tcSubType and check whether the constraints are satisfiable for possible types (as Simon talks about here https://mail.haskell.org/pipermail/ghc-devs/2015-February/008392.html). The current problem however is that `TcSimplify.hs` imports `TcErrors.hs`, so at very least I'd have to change some of the file structure if I'd like to proceed with also searching for types that are subsumable by the expected type.

### DIFF
--- a/compiler/main/DynFlags.hs
+++ b/compiler/main/DynFlags.hs
@@ -707,6 +707,8 @@ data DynFlags = DynFlags {
 
   maxRelevantBinds      :: Maybe Int,   -- ^ Maximum number of bindings from the type envt
                                         --   to show in type error messages
+  maxValidSubstitutions :: Maybe Int,   -- ^ Maximum number of substitutions
+                                        --   to show in type error messages
   maxUncoveredPatterns  :: Int,         -- ^ Maximum number of unmatched patterns to show
                                         --   in non-exhaustiveness warnings
   simplTickFactor       :: Int,         -- ^ Multiplier for simplifier ticks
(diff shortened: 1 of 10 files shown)


Tritlo:
Merged in https://github.com/ghc/ghc/commit/26c95f46e679fb73e1ec2bec2be0801c72fd1449

PR #30, opened 2017-04-02
Fix space leak in sortBy

greg7mdp:
Fix a space leak in sortBy, significantly increasing the performance of sorting random integers (about 30%) and strings (about 18%). Performance for sorting already sorted lists (both ascending and descending) is just slightly faster. Many thanks to Siddhanathan Shanmugam for helping me with this.

### DIFF
--- a/libraries/base/Data/OldList.hs
+++ b/libraries/base/Data/OldList.hs
@@ -854,14 +854,15 @@ sortBy cmp = mergeAll . sequences
 
     ascending a as (b:bs)
       | a `cmp` b /= GT = ascending b (\ys -> as (a:ys)) bs
-    ascending a as bs   = as [a]: sequences bs
+    ascending a as bs   = let x = as [a] in x `seq` x : sequences bs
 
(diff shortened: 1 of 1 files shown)


greg7mdp:
Anyone looking at merging this?

siddhanathan:
cc @ekmett I believe you are the maintainer for base?

mpickering:
You should submit the patch on phabricator so that they appear in usual work flows.

greg7mdp:
@mpickering , is the phabricator process really necessary for such a simple patch? At this point I am ready to give up.

mpickering:
I have put it up for you - https://phabricator.haskell.org/D3454

It isn't clear that this won't break performance tests (for better or worse) so it is better to test it with the CI infrastructure.

greg7mdp:
Great, thanks Matthew!

bgamari:
I'm going to close this in favor of https://phabricator.haskell.org/D3454.

PR #31, opened 2017-04-05
Modify behavior of -fno-code to generate enough code for template haskell to be happy 

duog:
Hi,
I submitted this to ticket [8025](https://ghc.haskell.org/trac/ghc/ticket/8025) last week, submitting it again here since I didn't get any replies.

I have not contributed to GHC before, and I would like to work on resolving this issue. Both haddock and flycheck should be able to use -fno-code to do a lot less work. Currently, haddock essentially turns off no-code when any modules use template haskell and flycheck uses no-link to work around both this issue and #10600.

I think this can be resolved by having ghc itself enable codegen on home modules depended on by modules that use template haskell.

downsweep is changed to do an additional pass over the modules, targetting any ModSummaries transitively depended on by a module that has LangExt.TemplateHaskell enabled. Those targeted modules have hscTarget changed from HscNothing to the default target of the platform.

There are two shell scripts in the root of my fork which boot and run an ad-hoc benchmark of cabal haddock on the wreq library. This library uses template haskell to generate lenses, and imports home modules from modules that use template haskell. However, most modules are not imported from modules that use template haskell. With this patch, cabal haddock runs in about half the time.

At least two points that need to be addressed:

Should this behaviour require an additional flag? It doesn't seem good for -fno-code to sometimes generate code.
Is the default target of the platform the correct choice?
I would require some guidance to get this into a mergable state, if you are interested in taking this fix please let me know the next step. I am happy to use your phabricator if that is appropriate.


bgamari:
Hi Duog, sorry for missing your Trac comment. I've left a [brief response](https://ghc.haskell.org/trac/ghc/ticket/8025#comment:8) and will look at your approach now. It would be great to get this up on Phabricator.

PR #33, opened 2017-04-20
Build system: use $(bindir)/{ghc-pkg,ghc} instead of  INSTALLED_GHC{,_PKG}_REAL

MaskRay:
`make install` from a binary-dist will fail if INSTALLED_GHC{,_PKG}_REAL are used:

```
$ utils/ghc-pkg/dist-install/build/tmp/ghc-pkg
...
error while loading shared libraries:
libHSterminfo-0.4.0.2-ghc8.3.20170419.so: cannot open shared object
file: No such file or directory
```


MaskRay:
Need more work.

PR #34, opened 2017-04-21
Build system: fix `INSTALLED_GHC{,_PKG}_REAL` when installing cross-compiled binary distribution on TARGET

MaskRay:
`make install` from a cross-compiled binary distribution will fail to run `$(ghc-cabal_INPLACE) register` in the rule `install_packages` if the old incorrect `INSTALLED_GHC{,_PKG}_REAL` are used:

```
$ utils/ghc-pkg/dist-install/build/tmp/ghc-pkg
...
error while loading shared libraries:
libHSterminfo-0.4.0.2-ghc8.3.20170419.so: cannot open shared object file: No such file or directory
```

### DIFF
--- a/ghc.mk
+++ b/ghc.mk
@@ -962,7 +962,7 @@ endif
 
 INSTALLED_PACKAGE_CONF=$(DESTDIR)$(topdir)/package.conf.d
 
-ifeq "$(CrossCompiling)" "YES"
+ifeq "$(BINDIST) $(CrossCompiling)" "NO YES"
 # when installing ghc-stage2 we can't run target's
 # 'ghc-pkg' and 'ghc-stage2' but those are needed for registration.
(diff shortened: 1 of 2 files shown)


MaskRay:
@trofi Refined.

Tested, `make install` works well on TARGET.

On BUILD:
```
% make show VALUE=BINDIST
make --no-print-directory -f ghc.mk show
BINDIST="no"
```

On TARGET:
```
~/tmp/ghc-8.3.20170421$ make show VALUE=BINDIST
make --no-print-directory -f ghc.mk show BINDIST=YES NO_INCLUDE_DEPS=YES
BINDIST="YES"
```

trofi:
Pushed slightly tweaked version as: https://github.com/ghc/ghc/commit/32a5ba993de4a8c88d4738d7386ba8afbaa84e78

Thank you!

bgamari:
Thanks @MaskRay!

PR #35, opened 2017-04-22
Don't describe tuple sections as "Python-style"

chris-martin:
I think this comparison introduces more confusion than clarity. Python, as far as I know, doesn't have any feature that resembles tuple sections. From [this discussion](https://www.reddit.com/r/programming/comments/93wwn/ghc_gets_pythonstyle_tuple_sections/) it seems like a bunch of other people were also confused by this when it came out.

### DIFF
--- a/docs/users_guide/glasgow_exts.rst
+++ b/docs/users_guide/glasgow_exts.rst
@@ -1570,7 +1570,7 @@ Tuple sections
 
     Allow the use of tuple section syntax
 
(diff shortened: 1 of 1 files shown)


bgamari:
Yep, I agree; the mention of Python only obscures the point.

bgamari:
Merged with 688272ba31df7bd0a094c3d86d60f7b77b9f5268.

PR #37, opened 2017-05-10
Fix crash in isModuleInterpreted for HsBoot (fixes #13591)

lspitzner:
Rename isModuleInterpreted to moduleIsBootOrNotObjectLinkable
because a) there already is a moduleIsInterpreted function in
the same module b) I have no idea if the (new) semantic of
the bool returned matches some understanding of "is interpreted".

This code now defaults to returning True for hs-boot files, which might
not be "correct". This choice was made due to the uses of this function:
It leads to sensible output in HscTypes:showModMsg and UI:modulesLoadedMsg.
The fix for #9887 (i.e. 41051dd) introduced (or at least exposed) this bug and
consequently it never worked for hs-boot.
This patch does not fix that; it just prevents the exception.

### DIFF
--- a/compiler/main/GHC.hs
+++ b/compiler/main/GHC.hs
@@ -112,7 +112,7 @@ module GHC (
         moduleIsInterpreted,
         getInfo,
         showModule,
-        isModuleInterpreted,
+        moduleIsBootOrNotObjectLinkable,
 
         -- ** Inspecting types and kinds
         exprType, TcRnExprMode(..),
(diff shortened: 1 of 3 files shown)


bgamari:
For the record, this is in my merge queue, which is a bit stalled due to unrelated issues.

bgamari:
This has been merged as 1edee7a8b5ca24156cb6e21bde6d611a0ba63882.

PR #39, opened 2017-05-18
Fix Haddock markup

vagarenko:
This fixes Haddock markup for `AppTypeE` constructor of `Exp` type.

Also markup for `UnboundVarE` constructor at line 1584
https://github.com/vagarenko/ghc/blob/0b876e1ec56459dcb79817b7760182d7a95acbd0/libraries/template-haskell/Language/Haskell/TH/Syntax.hs#L1584
looks suspicious. Should it be fixed same way?


niteria:
Merged with c039624, thanks!

You're right that that Haddock comment looks a bit off: https://hackage.haskell.org/package/template-haskell-2.11.1.0/docs/Language-Haskell-TH-Syntax.html#v:UnboundVarE

But I'm not sure how to fix it, the code block seems to always span till the end of the line. 
Perhaps putting `(hole)` on the next line would fix it. https://github.com/haskell/haddock/issues/447 seems related.

PR #41, opened 2017-05-22
Add missing "do" to example in arrow docs.

null-a:
I'm getting a parse error on this example without this change. After this change the example parses and compiles.


niteria:
Merged as 6f8c3ce, thanks!

PR #42, opened 2017-05-25
template-haskell: Properly escape StrTyLit doc

glguy:


niteria:
Merged as db1fd97, thanks!

PR #44, opened 2017-06-14
Add flag to hide the loaded modules in ghci

bitonic:
They're quite an eyesore when you have 200+ modules, but `v0` does not show progress.

### DIFF
--- a/compiler/main/DynFlags.hs
+++ b/compiler/main/DynFlags.hs
@@ -525,6 +525,7 @@ data GeneralFlag
    | Opt_OptimalApplicativeDo
    | Opt_VersionMacros
    | Opt_WholeArchiveHsLibs
+   | Opt_HideLoadedMods -- Hide lists of loaded modules in ghci
 
    -- PreInlining is on by default. The option is there just to see how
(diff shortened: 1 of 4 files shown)


bitonic:
Good idea, I scanned `docs/users_guide/ghci.rst` but I didn't find an obvious place, do you have a suggestion?

bgamari:
What you've added is merely a reference to a (non-existent) flag definition. See, for instance, line 1008 of `ghci.rst` for an example of how to document a flag.

simonmar:
Yeah, I believe we have a local hack in our custom GHCi to do this.  Perhaps we should even just elide the message by default, since it doesn't scale well.

(btw, I thought non-trivial pull requests were supposed to get moved to Phabricator?)

bgamari:
> (btw, I thought non-trivial pull requests were supposed to get moved to Phabricator?)

Indeed I was going to let @bitonic finish iterating here before moving this to Phabricator for a final review.

bgamari:
This has been superceded by [D3651](https://phabricator.haskell.org/D3651).

PR #46, opened 2017-06-17
Document mutable array copy overlap behavior

andrewthad:
From the current docs, it's not immediately obvious that copying from a mutable array to itself is allowed. You have to dig into the comments in the primop generator to figure this out. What makes matters worse is that `primitive` documents its wrapper functions (`copyMutableByteArray`, etc.) with:

> Copy a slice of a mutable byte array into another array. The two slices may not overlap.

I assume that this was true in older GHCs, but it's certainly not true anymore. The PR improves the haddocks for all the mutable array copy functions to mention that overlapping regions are supported.

### DIFF
--- a/compiler/prelude/primops.txt.pp
+++ b/compiler/prelude/primops.txt.pp
@@ -827,7 +827,9 @@ primop  CopyMutableArrayOp "copyMutableArray#" GenPrimOp
    number of elements to copy, copy the elements from the source array
    to the destination array. The source and destination arrays can
    refer to the same array. Both arrays must fully contain the
-   specified ranges, but this is not checked.}
+   specified ranges, but this is not checked. 
+   The regions are allowed to overlap, although this is only possible when the same 
+   array is provided as both the source and the destination. }
   with
   out_of_line      = True
   has_side_effects = True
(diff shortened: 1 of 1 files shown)


compiler/prelude/primops.txt.pp
```diff
@@ -990,7 +992,9 @@
    number of elements to copy, copy the elements from the source array
    to the destination array. The source and destination arrays can
    refer to the same array. Both arrays must fully contain the
-   specified ranges, but this is not checked.}
+   specified ranges, but this is not checked.
+   The regions are allowed to overlap, although this is only possible when the same 
+   array is provided as both the source and the destination. }
```
bgamari: I feel like this wording is a tad confusing. Perhaps,
> In the case where the source and destination are the same array the source and destination regions may overlap
I don't feel strong one way or another though.

andrewthad: Yeah, I prefer your wording. I'll change it to that.

andrewthad:
I've changed the wording to:

> In the case where the source and destination are the same array the source and destination regions may overlap

bgamari:
Thanks!

bgamari:
Merged

PR #51, opened 2017-07-13
improve the docs for finalizers to clarify when finalizers may not be run

andrewthad:
Adds documentation to `System.Mem.Weak` to clarify what guarantees users can expect about finalizers being run. This doesn't fix https://ghc.haskell.org/trac/ghc/ticket/13167, but it documents the problem.

### DIFF
--- a/libraries/base/System/Mem/Weak.hs
+++ b/libraries/base/System/Mem/Weak.hs
@@ -146,7 +146,7 @@ A heap object is /reachable/ if:
 
 {- $notes
 
-A finalizers is not always called after its weak pointer\'s object becomes
+A finalizer is not always called after its weak pointer\'s object becomes
 unreachable. There are two situations that can cause this:
 
  * If the object becomes unreachable right before the program exits,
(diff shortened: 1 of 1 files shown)


bgamari:
Thanks for doing this. It looks reasonable to me but I noticed a few typos that should be fixed.

andrewthad:
@bgamari I fixed the three typos I could find. I totally should have proofread before PRing :no_mouth:

bgamari:
Oh dear, it looks like I neglected to submit my review. Sorry about that!

libraries/base/System/Mem/Weak.hs
```diff
@@ -140,3 +144,25 @@ A heap object is /reachable/ if:
  * It is the value or finalizer of a weak pointer object whose key is reachable.
 -}
 
+{- $notes
+
+A finalizers is not always called after its weak pointer\'s object becomes
+unreachable. There are two situations that can cause this:
+
+ * If the object becomes unreachable right before the program exits,
+   then GC may not be performed. Finalizers run during GC, so finalizers
```
simonmar: Almost correct, in fact what happens is the GC runs and creates a thread for the finalizer, but the program exits before the thread runs.  (could reference the other docs about not waiting for threads to finish before program exit).

PR #53, opened 2017-08-01
Alpine Linux patches forward port (trac#14057)

tuncer:


tuncer:
https://ghc.haskell.org/trac/ghc/ticket/14057

bgamari:
I'm back from travelling now. I'll try to have a look at this shortly.

tuncer:
Okay.

tuncer:
Pushed minor update fixing punctuation of comments.

compiler/iface/MkIface.hs
```diff
@@ -713,6 +713,8 @@ addFingerprints hsc_env mb_old_fingerprint iface0 new_decls
     (non_orph_fis,   orph_fis)   = mkOrphMap ifFamInstOrph (mi_fam_insts iface0)
     fix_fn = mi_fix_fn iface0
     ann_fn = mkIfaceAnnCache (mi_anns iface0)
+    -- Do not allow filenames to affect the interface.
+    usages = [ case u of UsageFile _ fp -> UsageFile "" fp; _ -> u | u <- mi_usages iface0 ]
```
angerman: I find this hard to read. And I really got confused with the naming here. Especially as I read `fp` as file path, and not fingerprint.

How about
```hs
-- anonUsage drops the file path and retains the fingerprint from the UsageFile,
-- so that the path does not affect the interface.
anonUsage :: Usage -> Usage
anonUsage (UsageFile _path hash) = UsageFile "" hash
anonUsage u = u
```
and using `map anonUsage (mi_usages iface0)`?

angerman:
Can this PR be called "Upstream muslc compat patches from Apline Trac#14057" or something? And get some extended description?

compiler/iface/MkIface.hs
tuncer: It's a forward port of the original diff and I don't have a strong style preference in this case. If you send me a pull request for the branch in my fork, I'll merge it and it will show up here. Would that work for you?

tuncer:
@angerman wrote:
> In general this looks good to me. The $(SHELL) part should probably get some additional comment. Even though it's buried in one of the commits.

Where do you suggest to add more info? The commit message explains the rationale for the diff, and we could comment the shell script, but that doesn't seem like a good idea to me.

tuncer:
@angerman, thanks for the review, btw. The original plan was to get this into 8.2.2, and then inspect the rest of the diffs, which includes stuff that's not as evidently upstream worthy.

bgamari:
I have merged the first two of these to `master`. The other two are just non-trivial enough that I need to think a bit.

bgamari:
I'm tracking the remainder of these in [#14057](https://ghc.haskell.org/trac/ghc/ticket/14057#ticket). Let's continue this there.

PR #55, opened 2017-08-05
Fix string escaping in Json

dmalikov:
It seems that double quotes is not escaped properly at the moment.

We'd noticed this with @alexbiehl during the work on https://github.com/haskell/haddock/pull/645

### DIFF
--- a/compiler/utils/Json.hs
+++ b/compiler/utils/Json.hs
@@ -39,7 +39,7 @@ escapeJsonString = concatMap escapeChar
     escapeChar '\n' = "\\n"
     escapeChar '\r' = "\\r"
     escapeChar '\t' = "\\t"
-    escapeChar '"'  = "\""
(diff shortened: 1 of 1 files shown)


hvr:
that's not the only bug in that implementation... :-)

Shame-less plug: here's a "better" dependency-free mini implemetnation: https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Utils/Json.hs

dmalikov:
@hvr can we start at least with this fix? What should be done to have this in upstream HEAD?

bgamari:
Yes, this is at very least a start.

@hvr, feel free to open a diff replacing the JSON implementation if you'd like.

dmalikov:
@bgamari thank you for merging

Closing, since e8fe12f83b17dc39d9272d44c4168946fa54e7a0 is in upstream

PR #62, opened 2017-08-11
Fix Applicative doc code

quchen:
Due to a missing newline between the `@`, Haddock does not render the two-line example correctly. Inserting a blank line should fix the issue.

> selection_003


bgamari:
Good catch!

bgamari:
Merged.

PR #72, opened 2017-08-29
Typo Fix `necssary`

h4ck3rm1k3:
http://ghc.haskell.org/trac/ghc/ticket/14166


h4ck3rm1k3:
http://ghc.haskell.org/trac/ghc/ticket/14166

bgamari:
Thanks for catching this! Usually we try to batch these sorts of simple typo fixes so our revision history isn't cluttered with lots of tiny patches which need to be bisected through later. Nevertheless, given that this is the only mistake you found we can merge this as-is.

Thanks again!

bgamari:
Merged as 156152548fb7a05317a7b97c45dd0437b57a26b3. Thanks again!

PR #74, opened 2017-09-07
Add non-ASCII isLetter True example

chris-martin:
I didn't find it immediately obvious from the documentation whether `isLetter` could produce `True` for any non-ASCII characters - it seemed plausible that only `[a-zA-Z]` would count. I figure one more example just to make this explicit can't hurt.


bgamari:
Sure, why not.

bgamari:
Merged. Thanks, @chris-martin!

PR #78, opened 2017-10-26
Data.Maybe: generalize catMaybes

phunehehe:
This seems like an innocuous change: the test results are the same as before, and `catMaybes` is slightly better. It now works with other list-like structures. Is this kind of change wanted?


ekmett:
One headache is that a slightly different, but incompatible, generalization is available through `Foldable`, which requires much less of `f`, but doesn't reconstruct it in the same form.

```
catMaybes :: Foldable f => f (Maybe a) -> [a]
catMaybes = foldMap toList
```

bgamari:
I agree that this sort of thing really needs to be through the Core Libraries Committee. Do let us know if you need help getting started with this, @phunehehe.

PR #80, opened 2017-10-31
change example from msum to mfilter

GinBaby:
Since `msum` does not do what I think most people intuitively understand as *summing*, I thought the example in the naming conventions documentation was a little misleading. I thought perhaps changing that example to contrast `filter` and `mfilter` would be more illustrative since there is not a discrepancy in the (probably, intuitively) expected result.


bgamari:
Sure, why not.

bgamari:
Merged with df479f7c1e3f877e9981f16374bfdf03e79e57f4.

PR #81, opened 2017-11-01
change bifunctor documentation; add examples

GinBaby:
I thought the `Bifunctor` documentation could use a little expansion and addition of some examples.

### DIFF
--- a/libraries/base/Data/Bifunctor.hs
+++ b/libraries/base/Data/Bifunctor.hs
@@ -20,11 +20,18 @@ module Data.Bifunctor
 import Control.Applicative  ( Const(..) )
 import GHC.Generics ( K1(..) )
 
--- | Formally, the class 'Bifunctor' represents a bifunctor
+-- | A bifunctor is a type constructor that takes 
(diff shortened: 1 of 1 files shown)


libraries/base/Data/Bifunctor.hs
```diff
@@ -59,22 +67,49 @@ class Bifunctor p where
...
+    -- Left 'J'
     first :: (a -> b) -> p a c -> p b c
     first f = bimap f id
 
+
     -- | Map covariantly over the second argument.
     --
     -- @'second' ≡ 'bimap' 'id'@
+    --
+    --==== Examples:
```
bgamari: Do you know whether a space is required here? I'm a bit worried that haddock will get confused here.

GinBaby: I also do not know whether it's required, but I've fixed my PR to match the other examples of it I could find in `base` (e.g., in Data.List). Sorry for confusion.

bgamari:
Thanks!

bgamari:
Merged with 275ac8ef0a0081f16abbfb8934e10cf271573768.

PR #83, opened 2017-11-03
CircleCI setup for Linux and macOS

mchakravarty:
* Also includes some support for FreeBSD (but blocked on FreeBSD failures right now).
* Also includes preliminary Windows support for AppVeyor (but this is still in testing)


bgamari:
I'm a bit confused; GitHub seems to think that the CircleCI builds were cancelled yet they look quite complete to me.

bgamari:
Ahh, I see: CircleCI says "There was an issue while running this container and it was rerun. The most recent run is shown."

I'm also quite confused by the fact that when I attempt to download the log (https://circleci.com/api/v1.1/project/github/tweag/ghc/170/output/104/0?file=true) my browser (and `curl`, for that matter) claim that the log is 0 bytes. This, despite the fact that the web interface claims that the log is over 400000 characters long.

bgamari:
Alright, quirks aside, it would be nice if the `make test` output were separate from the build output.

mboes:
You seem to be using the 1.1 endpoint, when in fact this should be running on the 2.0 infrastructure. Maybe that's why you have an issue with the logs?

> it would be nice if the make test output were separate from the build output.

Sounds like a reasonable nice-to-have. But is this PR as it stands functional enough to merge?

ghc.mk
```diff
@@ -1091,7 +1091,7 @@ $(eval $(call bindist-list,.,\
     $(wildcard libraries/*/*/dist-install/doc/) \
     $(filter-out settings llvm-targets,$(INSTALL_LIBS)) \
     $(RTS_INSTALL_LIBS) \
-    $(filter-out %/project.mk mk/config.mk %/mk/install.mk,$(MAKEFILE_LIST)) \
+    $(filter-out testsuite/utils/check-ppr/% testsuite/utils/check-api-annotations/% %/project.mk mk/config.mk %/mk/install.mk,$(MAKEFILE_LIST)) \
```
bgamari: Presumably this shouldn't be here?

.circleci/prepare-system.sh
```diff
@@ -0,0 +1,46 @@
...
+}
+
+case "$(uname)" in
+  Linux)
+    if [[ -n ${TARGET:-} ]]; then
+      if [[ $TARGET = FreeBSD ]]; then
+        # cross-compiling to FreeBSD
+        add-apt-repository -y ppa:hvr/ghc
+        apt-get update -qq
+        apt-get install -qy ghc-8.0.2 cabal-install  alex happy ncurses-dev git openssh-client make automake autoconf gcc perl python3 texinfo xz-utils
```
hvr: Note that `cabal-install` does not exist in my ppa, instead it will result in a likely quite old `cabal` to be installed; my PPA provides `cabal-install` packages suffixed by the major version, e.g. `cabal-install-2.0` or `cabal-install-1.24`, or even `cabal-install-head`.

mchakravarty: @hvr Thanks for pointing that out. Which version would you suggest to use here with GHC 8.0.2?

ghc.mk
Fuuzetsu: Not anymore: the revert in d9b6015d1942aa176e85bb71f34200bab54e1c9c makes it unnecessary, will remove this commit.

.circleci/prepare-system.sh
Fuuzetsu: > GHC 8.0 only works with cabal-install version 1.24 or later. Please upgrade if you have an older version of cabal-install. 

We'll go with 1.24 for now.

Fuuzetsu:
@bgamari @hvr I have addressed all the comments, PATAL.

Fuuzetsu:
I'm unsure why CircleCI is stuck in Running for OSX job: everything went fine, it just refuses to terminate. I have submitted a support ticket to CircleCI in meantime: possibly we may want to make the Darwin check optional for now.

bgamari:
Huh, odd. Stuck uploading artifacts. Alright, do keep us informed about the status of the ticket. Thanks again for your work on this, @mchakravarty  and @Fuuzetsu. I'll merge shortly.

mchakravarty:
> Huh, odd. Stuck uploading artifacts. Alright, do keep us informed about the status of the ticket. Thanks again for your work on this, @mchakravarty and @Fuuzetsu. I'll merge shortly.

Great, thanks everybody!

bgamari:
The OS X situation still hasn't stabilized it seems. I've opened [#14448](https://ghc.haskell.org/trac/ghc/ticket/14448) so we don't lose track of this.

PR #87, opened 2017-11-13
make documentation of atomically more accurate

andrewthad:
I started a [discussion](https://mail.haskell.org/pipermail/libraries/2017-November/028300.html) on the libraries mailing list about this. Basically, the docs for `atomically` are currently misleading. I noticed this because I in a project I'm working on, I wanted to use `atomically` inside `unsafePerformIO`. I discovered that it did not throw an exception as I expected it to. In this PR, I've updated the docs to reflect, to the best of my understanding, the expectations a user can have when using `atomically` in `unsafePerformIO`.

### DIFF
--- a/libraries/base/GHC/Conc/Sync.hs
+++ b/libraries/base/GHC/Conc/Sync.hs
@@ -716,13 +716,17 @@ unsafeIOToSTM (IO m) = STM m
 
 -- | Perform a series of STM actions atomically.
 --
--- You cannot use 'atomically' inside an 'unsafePerformIO' or 'unsafeInterleaveIO'.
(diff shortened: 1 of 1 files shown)


andrewthad:
I need to add something about the pitfalls of using unsafePerformIO inside of atomically as well. I could probably just link to unsafeIoToStm since the dangers are basically the same.

andrewthad:
@bgamari I've pushed more changes that address the problem with making `newTVarIO` sound unique. Also, I've added a brief mention that `unsafePerformIO` inside of `atomically` can be bad too.

bgamari:
Thanks @andrewthad! Pushed.

PR #90, opened 2017-11-30
Only look for locales of the form LL.VV

ggreif:
because in recent RHEL7 suddenly locales like `bokmål` pop up, which screw up reading-in of ASCII strings a line later. This additional criterion reliably eliminates those unicode characters.


bgamari:
I suppose this is reasonable. Sure.

ggreif:
It ended up in master via Phab.

PR #93, opened 2017-12-16
Prevent "C--" translating to "C<endash>" in the User's Guide.

steshaw:
i.e. escape "--" so that it doesn't translate to an "en dash" character. I did this by providing a `|Cmm|` substitution.

Note: I would have preferred to add the `|Cmm|` substitution in a single place — either `conf.py` in the existing `rst_prolog`, `rst_epilog`, or alternatively an included file. However, each of these options provoked a runtime exception from docutils (on Sphinx 1.6.5). Therefore, I've defined `|Cmm|` at the bottom of the 3 affected files.

### DIFF
--- a/docs/users_guide/debug-info.rst
+++ b/docs/users_guide/debug-info.rst
@@ -268,10 +268,10 @@ GHC may produce the following standard DIEs in the ``.debug_info`` section,
   Represents a compilation unit (e.g. a Haskell module).
 
 ``DW_TAG_subprogram``
-  Represents a C-- top-level basic block.
+  Represents a C-\\- top-level basic block.
 
 ``DW_TAG_lexical_block``
-  Represents a C-- basic block. Note that this is a slight departure from the
+  Represents a C-\\- basic block. Note that this is a slight departure from the
   intended meaning of this DIE type as it does not necessarily reflect
   lexical scope in the source program.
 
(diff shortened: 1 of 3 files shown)


bgamari:
Thanks for taking the time to fix this!

ggreif:
Cool!

bgamari:
Unfortunately this seems to break the manpage build:
```
debugging:1: WARNING: "substitution_reference" not supported                   

Exception occurred:
  File "/usr/lib/python2.7/dist-packages/docutils/nodes.py", line 1919, in unknown_departure
    % (self.__class__, node.__class__.__name__))
NotImplementedError: sphinx.writers.manpage.ManualPageTranslator departing unknown node type: substitution_reference
```
It's not entirely clear what can be done about this. Even if we could fix the apparent bug in Sphinx, we still wouldn't be able to add this for quite some time due to compatibility. I'm afraid we'll need to find another way around this.

steshaw:
That looks like the exception I was getting when I tried to put the `|Cmm|` substitution definition in a single place.

I could certainly inline the definition of the substitution (it's just slightly ugly and I though best not repeated often). However, perhaps I can find some other workaround. How do I do a manpage build?

nomeata:
> I could certainly inline the definition of the substitution

It’s not the worst. It looks a bit ugly, but it is easier to understand for a drive-by contributor, and they learn something about restful syntax along the way.

steshaw:
No worries!

I've inlined the `|Cmm|` substitution and force pushed. AFAICT, this now builds the user's guide _and_ the man page. To check, I ran the following in `docs/users_guide`:

```sh-session
$ make all build-man/ghc.1
```

Previously, I was only doing a `make` in `docs/users_guide`.

bgamari:
This was merged with 12f5c00543e1f3dc7109e3575fcc4a973aebdacc.

PR #95, opened 2018-01-16
Derive more Data methods

treeowl:
Summary:
* Derive more `Data` methods. I'm hoping that this will save compilation
  time by starting with the code we want rather than having to inline
  `gfoldl` everywhere and reduce. While I'm currently focused on improving
  the amount of time it takes to compile the *instances*, I expect this
  will also reduce the amount of time necessary to compile code using them.
  For example, `gmapM` should start out with the monadic binds associated
  the right way to begin with, rather than making GHC reassociate them all.
  I have not yet written code to derive `gmapMp` or `gmapMo`; let's first
  see what happens to the rest. I've also left `gmapQ` alone for now based
  on a guess that it should be really easy to compile well--inline `gmapQr`
  once and everything should be perfect.

* Swap the arguments to `zip` in `mkConstr` to allow the enumeration
  to fuse away.

Reviewers: hvr, bgamari

Subscribers: rwbarton, thomie, carter

GHC Trac Issues: #14594

Differential Revision: https://phabricator.haskell.org/D4318

### DIFF
--- a/compiler/typecheck/TcGenDeriv.hs
+++ b/compiler/typecheck/TcGenDeriv.hs
@@ -1271,7 +1271,7 @@ we generate
   instance (Data a, Data b) => Data (T a b) where
     gfoldl k z (T1 a b) = z T `k` a `k` b
     gfoldl k z T2           = z T2
-    -- ToDo: add gmapT,Q,M, gfoldr
+    -- ToDo: add Q, gfoldr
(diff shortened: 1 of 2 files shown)


bgamari:
What is the status of this?

treeowl:
Not ready.

PR #97, opened 2018-01-18
Invert likeliness when improving conditionals

alexbiehl:
... in CmmSink


AndreasPK:
How about defining inv_likeliness as:

inv_likeliness = fmap not

PR #98, opened 2018-01-30
Improve try remove unused syntax overhead

saschagrunert:


libraries/base/Control/Exception/Base.hs
```diff
@@ -172,7 +172,7 @@ mapException f v = unsafePerformIO (catch (evaluate v)
 -- >  try a = catch (Right `liftM` a) (return . Left)
 
 try :: Exception e => IO a -> IO (Either e a)
-try a = catch (a >>= \ v -> return (Right v)) (\e -> return (Left e))
+try a = catch (a >>= \v -> return (Right v)) (return . Left)
```
int-index: How does changing a lambda to composition affect the inliner behavior in this case?

mpickering: I think it's probably fine but I don't see why this style is better than using an explicit lambda anyway.

alexbiehl: I think it doesn't matter here. `catch` is implemented through the `catch#` primop. Inlining can't see through `catch#`. So GHC can't inline the `(return . Left)` part and will compile it to a static closure instead.

mpickering: but `(.)` will get inlined so the result will be the same as the eta-expanded version yes?

alexbiehl: Ah, yes! Sorry for the confusion. What I wanted to add to your statement is that `return . Left` won't be simplified in a context outside this very lambda.

saschagrunert: It is just my personal opinion that it looks more sexy :+1: Some linters like hlint will also "complain" like "avoid lambda..."

bgamari:
While I do appreciate the thought, I really don't see any compelling benefits to merging this. The difference is purely stylistic and in this case I don't think the explicit nature of the current implementation to a bad thing; to the contrary, it makes it obvious how the code will behave when compiled.

PR #112, opened 2018-03-19
Full AppVeyor build with tests

mrkkrp:
We have succeeded in running full build with tests on AppVeyor.

Times:

* Only build (default build flavor): 1h 53min.
* Build with fasttests: 2h 15min.
* Build with tests: 2h 28min (this PR).

This is virtually the same times we've got for `validate-x86_64-linux-llvm` job in Circle CI, so I guess it's an acceptable result.

Just merging this PR is not sufficient of course because it's necessary to setup things in your AppVeyor account so it'll use private build cloud we've got now with image I've prepared. If you add me as an admin to GHC AppVeyor account I can handle all of this.

Some tests fail: https://ci.appveyor.com/project/mboes/ghc/build/129, but that's another story.

@bgamari Let me know what you think.

### DIFF
--- a/.appveyor.sh
+++ b/.appveyor.sh
@@ -1,8 +1,8 @@
 # Configure the environment
 MSYSTEM=MINGW64
-THREADS=3
+THREADS=9
 SKIP_PERF_TESTS=YES
-BUILD_FLAVOUR=quick
+BUILD_FLAVOUR=
 source /etc/profile || true # a terrible, terrible workaround for msys2 brokenness
 
 # Don't set -e until after /etc/profile is sourced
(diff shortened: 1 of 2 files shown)


bgamari:
This looks quite reasonable to me. Great work, @mrkkrp!

I believe you should now be added as an administrator of the `GHCAppveyor` team. Do let me know if you need anything else.

mrkkrp:
Full build now works, this may be merged. I ended up even fixing that `Stats.c` & `Stats.h` thing, because the mirror apparently was not yet updated and my builds were failing.

Rebased against current master.

Example of full build: https://ci.appveyor.com/project/GHCAppveyor/ghc/build/315.

bgamari:
Merged.

PR #114, opened 2018-03-22
Add a job running on Fedora

mrkkrp:
This adds a job running on Fedora 26. Trac ticket: https://ghc.haskell.org/trac/ghc/ticket/14949.

Can you clarify what you mean by

> In principle we only need to produce one bindist per combination

Do you mean that we should add one job per line in [this table](https://ghc.haskell.org/trac/ghc/wiki/Platforms/Linux), but if so, maybe it's too many? If "per combination" means that we we should add even more jobs, then I'm not sure it's a good idea.

Currently I see that bindists include the following:

OS | Job
----|-----
Linux (x86) | `validate-i386-linux`
Linux (x64_64) | `validate-x86_64-linux`
Linux (ARMv7) | ?
Linux (AArch64) | ?
Windows (x86) | ?
Windows (x86_64) | AppVeyor build
Mac OS X (x86_64) | `validate-x86_64-darwin`

[Taken from here](https://www.haskell.org/ghc/download_ghc_8_4_1.html). So I don't understand why we want to generate artifacts on Fedora. GHC built on Debian worked there fine. I'd expect that we need to add jobs that are missing, i.e. that are marked with ? in the table above if we're to generate all bindists on CI.

### DIFF
--- a/.circleci/images/x86_64-linux-fedora/Dockerfile
+++ b/.circleci/images/x86_64-linux-fedora/Dockerfile
@@ -1,0 +1,26 @@
+FROM fedora:27
+
+ENV LANG C.UTF-8
+
+RUN dnf -y install coreutils binutils which git make automake autoconf gcc perl python3 texinfo xz lbzip2 patch openssh-clients sudo curl zlib-devel sqlite ncurses-compat-libs gmp-devel ncurses-devel gcc-c++ findutils
+
+# Install GHC and cabal
+RUN cd /tmp && curl https://downloads.haskell.org/~ghc/8.2.2/ghc-8.2.2-x86_64-deb8-linux.tar.xz | tar -Jx
+RUN cd /tmp/ghc-8.2.2 && ./configure --prefix=/opt/ghc/8.2.2
+RUN cd /tmp/ghc-8.2.2 && make install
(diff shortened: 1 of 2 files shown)


bgamari:
The Fedora 27 build is intended for distributions using `libncurses.so.6`. However, it occurred to me as I was preparing the [distributions table](https://ghc.haskell.org/trac/ghc/wiki/Platforms/Linux) that we could prepare this distribution on Debian 9 instead.

> Do you mean that we should add one job per line in this table, but if so, maybe it's too many? If "per combination" means that we we should add even more jobs, then I'm not sure it's a good idea.

I mean we only need one job per combination of (gmp version, ncurses version).

.circleci/images/x86_64-linux-fedora/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM fedora:26
```
bgamari: This needs to be Fedora 27 as the goal is to support ncurses 6.

Alternatively we could use Debian Stretch.

mrkkrp:
Adjusted to use Fedora 27.

bgamari:
Thanks again, @mrkkrp!

PR #117, opened 2018-03-30
Collect artifacts with s3

mrkkrp:
This PR allows to store GHC artifacts for nightly builds (commits on `master`) and releases (tags starting with `ghc-`). Here is a spec (and source code of the helper): https://github.com/tweag/ghc-artifact-collector#behavior-spec. Please let me know if changes in behavior are desirable.

This link allows to see what's currently in the bucket: https://ghc-artifacts.s3.amazonaws.com/.

I've set environment variables for AppVeyor account, where I'm admin, but I have no permissions to configure Circle CI. If you can make an admin there too, that'd be great, otherwise I can send you the values of the variables and you'll have to set them yourself.

Currently I uploaded patched versions of the helper that will upload things generated on this branch `collect-artifacts-with-s3`. Once this is confirmed to work, I'll remove the commit that disables the tests (they'd prevent artifact uploading) and upload proper versions of the tool that stores things only on `master`. Next we'll need to make sure that Circle CI and AppVeyor won't set secret environment variables for PRs opened not from the same repo, this will prevent exploiting of the credentials.

cc @bgamari

### DIFF
--- a/.appveyor.sh
+++ b/.appveyor.sh
@@ -33,7 +33,8 @@ EOF
     "test")
         make test THREADS=$THREADS
         make binary-dist
-        7z a ghc-windows.zip *.tar.xz
+        curl https://ghc-artifacts.s3.amazonaws.com/tools/ghc-artifact-collector-x86_64-windows --output ghc-artifact-collector
+        ./ghc-artifact-collector *.tar.xz
         ;;
 
     *)
--- a/.appveyor.sh
+++ b/.appveyor.sh
@@ -31,10 +31,10 @@ EOF
         ;;
 
     "test")
-        make test THREADS=$THREADS
         make binary-dist
         curl https://ghc-artifacts.s3.amazonaws.com/tools/ghc-artifact-collector-x86_64-windows --output ghc-artifact-collector
         ./ghc-artifact-collector *.tar.xz
+        make test THREADS=$THREADS
         ;;
 
     *)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,9 +96,9 @@ jobs:
       - *boot
       - *configure_unix
       - *make
-      - *test
       - *bindist
       - *storeartifacts
+      - *test
 
   "validate-x86_64-freebsd":
     resource_class: xlarge
@@ -115,9 +115,9 @@ jobs:
       - *boot
       - *configure_bsd
       - *make
-      - *test
       - *bindist
       - *storeartifacts
+      - *test
 
   "validate-x86_64-darwin":
     macos:
@@ -136,9 +136,9 @@ jobs:
       - *boot
       - *configure_unix
       - *make
-      - *test
       - *bindist
       - *storeartifacts
+      - *test
 
   "validate-hadrian-x86_64-linux":
     resource_class: xlarge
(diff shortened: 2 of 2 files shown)


mrkkrp:
This works. Removing the commit that disables the tests...

mrkkrp:
All done. Pending: setting env vars for Circle CI.

mrkkrp:
@bgamari Ping. This is ready. In latest commit I moved artifact uploading before running tests. Commit message clarifies why:

> Since for every job there are failing tests, if we try to upload artifacts after tests, we won't get any. As a temporary measure I suggest uploading artifacts before running tests. Once tests are fixed, we should perform artifact uploading after tests again.

I hope it's OK for now. I'd like to have some artifacts for Stackage HEAD project, because recompiling GHC there looks like a waste of time to me if we store these artifacts anyway.

.appveyor.sh
```diff
@@ -31,9 +31,10 @@ EOF
         ;;
 
     "test")
-        make test THREADS=$THREADS
         make binary-dist
-        7z a ghc-windows.zip *.tar.xz
+        curl https://ghc-artifacts.s3.amazonaws.com/tools/ghc-artifact-collector-x86_64-windows --output ghc-artifact-collector
```
bgamari: Is this tool tracked under version control somewhere? Is there any reason we can't just include it in the GHC repo?

.circleci/config.yml
```diff
@@ -90,17 +88,17 @@ jobs:
...
       - *prepare
       - *submodules
       - *boot
       - *configure_unix
       - *make
-      - *test
       - *bindist
-      - *collectartifacts
       - *storeartifacts
+      - *test
```
bgamari: It would be nice if we could record the testsuite results alongside the artifact for posterity. That being said, this is fine for now.

.appveyor.sh
mboes: Agreed. I think we should eventually rebuild from source, for hermeticity and hence reproducibility. Shouldn't be a blocker though.

mrkkrp: The repo is here: https://github.com/tweag/ghc-artifact-collector. It has its own CI and commits on `master` are deployed (i.e. uploaded to S3) automatically. I think it may be handier to have this project in a separate repository because it's easier for me to hack on it and see changes almost instantly this way. If it's in GHC repo I'll be working through PRs and it's much much slower. We could include it in GHC repo when the helper stabilizes.

mrkkrp:
@bgamari Don't forget, we need to setup env vars for CircleCI for this to work:

> I've set environment variables for AppVeyor account, where I'm admin, but I have no permissions to configure Circle CI. If you can make an admin there too, that'd be great, otherwise I can send you the values of the variables and you'll have to set them yourself.

.appveyor.sh
mboes: The following two questions are orthogonal:

1. do we move it in-repo or keep it out?
2. do we use a prebuilt version pulled from somewhere?

The latter is independent of the former. If you push a prebuilt version somewhere and download it, you make it harder for others to fork (consider what happens if people want to try their own `ghc-artifcats-collector`). It's also not a hermetic setup.

PR #122, opened 2018-04-10
Add a wrapped applicative type, 'Ap', to Data.Monoid

chessai:
This is a small P.R., so I thought that I should make it here.

The idea and usefulness of this type have been brought up multiple times on the libraries mailing list. Everyone involved seems to be in favour of its inclusion, however no one could decide on a name at any point in time. I think 'Ap' is a good name, but I would ask that the maintainers of base (@ekmett, @RyanGlScott are the only two of you, as far as I am aware) ultimately decide on the name. If anything is wrong with this P.R., please let me know.

P.S. If possible, I would also like to make some inclusions in Data.Foldable (not to the typeclass) that are related to the use of this type, such as foldlMap and foldrMap. Please let me know if that is okay.


RyanGlScott:
Before we can even discuss this PR, it looks like something has gone horribly wrong—GitHub claims that 19,548 files are going to be changed. Can you correct this?

chessai:
@RyanGlScott sorry about that. I forgot to ban a ghc I had built. The issue seems to be corrected now.

andrewthad:
This should probably be rebased and force pushed.

23Skidoo:
As a minimum, this needs better docs.

chessai:
closing in favour of #123

ekmett:
@bgamari: We're pretty content to just have this done. =) Moving further discussion to #123

PR #126, opened 2018-04-20
improve documentation of indexArray#

andrewthad:
For a long time, I had trouble understanding why the result of this was in an unboxed tuple. The monadic index functions in `primitive` and `vector` obscured its meaning even more for me. This is an attempt at explaining how the promptness of the pattern match on the unboxed tuple can affect people's code.

### DIFF
--- a/compiler/prelude/primops.txt.pp
+++ b/compiler/prelude/primops.txt.pp
@@ -818,7 +818,7 @@ primop  SizeofMutableArrayOp "sizeofMutableArray#" GenPrimOp
 
 primop  IndexArrayOp "indexArray#" GenPrimOp
    Array# a -> Int# -> (# a #)
-   {Read from specified index of immutable array. Result is packaged
+   {Read from the specified index of an immutable array. The result is packaged
     into an unboxed unary tuple; the result itself is not yet
     evaluated. Pattern matching on the tuple forces the indexing of the
     array to happen but does not evaluate the element itself. Evaluating
--- a/compiler/prelude/primops.txt.pp
+++ b/compiler/prelude/primops.txt.pp
@@ -818,14 +818,13 @@ primop  SizeofMutableArrayOp "sizeofMutableArray#" GenPrimOp
 
 primop  IndexArrayOp "indexArray#" GenPrimOp
    Array# a -> Int# -> (# a #)
-   {Read from specified index of immutable array. Result is packaged into
-    an unboxed unary tuple; the result itself is not yet evaluated. Pattern
-    matching on the tuple forces the indexing of the array
-    to happen but does not evaluate the element itself. Doing this
-    pattern match immidiately can be useful for (1) preventing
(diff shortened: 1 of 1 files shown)


compiler/prelude/primops.txt.pp
```diff
@@ -819,7 +819,13 @@
 primop  IndexArrayOp "indexArray#" GenPrimOp
    Array# a -> Int# -> (# a #)
    {Read from specified index of immutable array. Result is packaged into
-    an unboxed singleton; the result itself is not yet evaluated.}
+    an unboxed unary tuple; the result itself is not yet evaluated. Pattern
+    matching on the tuple forces the indexing of the array
+    to happen but does not evaluate the element itself. Doing this
+    pattern match immidiately can be useful for (1) preventing
```
treeowl: s/immidiately/immediately

compiler/prelude/primops.txt.pp
```diff
@@ -819,7 +819,13 @@
...
+    additional thunks from building up on the heap and (2) eliminating
+    references to the argument array, allowing it to be garbage
+    collected more promptly.}
```
treeowl: I'd break this up into two sentences instead of using 1 and 2. I don't think referencing pattern matching on the tuple is really quite right, because it's unboxed.

andrewthad: Or I guess alternatively "forcing" or "evaluating".

treeowl:
"Evaluating" sounds good to me.

compiler/prelude/primops.txt.pp
```diff
@@ -818,8 +818,13 @@
 
 primop  IndexArrayOp "indexArray#" GenPrimOp
    Array# a -> Int# -> (# a #)
-   {Read from specified index of immutable array. Result is packaged into
-    an unboxed singleton; the result itself is not yet evaluated.}
+   {Read from specified index of immutable array. Result is packaged
+    into an unboxed unary tuple; the result itself is not yet
+    evaluated. Pattern matching on the tuple forces the indexing of the
+    array to happen but does not evaluate the element itself. Evaluating
+    the thunk prevents additional thunks from building up on the
```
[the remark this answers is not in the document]
andrewthad: But that's not right either. It's the `{\tt (# a #)}`. Would writing it like that show up correctly. Also, perhaps I should mention how an unboxed unary tuple is evaluated, since it's an uncommon construct. Something like:

> Evaluating the `(# a #)` (by casing on it or using a let binding to extract the inner value) prevents ...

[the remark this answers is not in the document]
andrewthad: The word "extract" is awful though. Maybe something more like "bind an identifier to the inner value".

bgamari:
This was merged in 15ece727d0cce777880ec2e3a1b3061f35dfce2c.

PR #129, opened 2018-04-26
Deprecate WrappedMonad

andrewthad:
This resolves https://ghc.haskell.org/trac/ghc/ticket/15027. This change has not been explicitly approved on trac or a mailing list, but I suspect utter uselessness of `WrappedMonad` compelled me to go ahead and put this up.

cc @RyanGlScott @ekmett

### DIFF
--- a/libraries/base/Control/Applicative.hs
+++ b/libraries/base/Control/Applicative.hs
@@ -63,11 +63,18 @@ import GHC.List (repeat, zipWith, drop)
 import GHC.Read (Read)
 import GHC.Show (Show)
 
+-- | Recover an 'Applicative' instance from a 'Monad' instance.
+--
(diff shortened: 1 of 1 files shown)


treeowl:
I think this needs approval from the libraries list. I think we do want to get rid of it; the main question is when.

treeowl:
Actually... I think this depends heavily on the `Monad` of no `return` proposal. If `return` sticks around in the long term, it *could* be useful to use `WrappedMonad` with `DerivingVia`. I haven't seen any final conclusion about MRP. Anyone know what its status is?

bgamari:
If anyone knows @hvr does.

bgamari:
Pinging @hvr.

PR #130, opened 2018-04-26
Update docker images to use GHC 8.4.2 and cabal-install-2.2

mrkkrp:
The `haskell` images are not being updated and there is no image with GHC 8.4.2, so we probably should not use them anymore. I adapted Dockerfiles used by those images so the end result is almost the same except we get newer GHC and cabal-install.

### DIFF
--- a/.circleci/images/x86_64-linux/Dockerfile
+++ b/.circleci/images/x86_64-linux/Dockerfile
@@ -1,10 +1,23 @@
-FROM haskell:8.2
+FROM debian:jessie
 
-# Make sure we have proper openssh before checkout: CircleCI git
-# does not check the repository out properly without it and also
-# takes 20 times longer than it should be.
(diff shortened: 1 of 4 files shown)


.circleci/images/x86_64-linux/Dockerfile
```diff
@@ -1,10 +1,23 @@
-FROM haskell:8.2
+FROM debian:jessie
```
hvr: Btw, if you use debian:stretch, you can make use of http://downloads.haskell.org/debian/

bgamari:
Yes, if the `haskell` images are abandoned then I agree we have to drop them. Thanks for doing this!

bgamari:
Merged in 56e8c6fc9d65fec11485bd02d437a753bc1a4ebe.

PR #131, opened 2018-04-26
Fix typo in user guide about promoted list

msakai:
`:'` in explanation text should be `':` to match code example.


bgamari:
Good catch!

bgamari:
Merged in .13e8bc0d7afa74d502a630c817840d390c328f6e.

PR #132, opened 2018-04-30
Use unsafeInsertNew to create timers in TimerManager

mitchellwrosen:
Timers are inserted into a priority queue with `insert`, which first calls `delete`. But our timeout keys are pulled from a unique supply, so we can skip the `delete` and call `unsafeInsertNew` instead. 

cc @jaspervdj - the PSQ code looks awfully similar to your `psqueues` package :)

### DIFF
--- a/libraries/base/GHC/Event/PSQ.hs
+++ b/libraries/base/GHC/Event/PSQ.hs
@@ -28,7 +28,7 @@ module GHC.Event.PSQ
     , singleton
 
     -- * Insertion
-    , insert
+    , unsafeInsertNew
 
     -- * Delete/Update
     , delete
@@ -36,7 +36,6 @@ module GHC.Event.PSQ
 
     -- * Conversion
     , toList
-    , fromList
 
     -- * Min
     , findMin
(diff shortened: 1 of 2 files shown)


treeowl:
Looks good to me. But please add a comment explaining why this is safe to make sure no one accidentally violates the invariant if they change something.

mitchellwrosen:
Good idea, done

libraries/base/GHC/Event/PSQ.hs
```diff
@@ -28,7 +28,7 @@ module GHC.Event.PSQ
     , singleton
 
     -- * Insertion
-    , insert
+    , unsafeInsertNew
```
bgamari: If `insert` isn't used then presumably we can drop its binding.

bgamari: Also, it would be a good idea to update the comment on `unsafeInsertNew` since it is no longer internal.

bgamari:
Thanks!

bgamari:
Merged with 418881f7181cbfa31c44f0794db65bf00916bde2.

PR #147, opened 2018-06-14
Fix documentation for `-dth-dec-file`

ntc2:
The docs said that it took a file argument, as in
`-dth-dec-file=<file>`, but it does not take a file argument!


bgamari:
Good catch!

bgamari:
Merged with 69954a125d12a1ffffcc9ce2d66bc211e3607a9a.

PR #154, opened 2018-06-28
Add "make run-ghci", which runs a ghci that can load compiler + ghci

mgsloan:
Major credit to Csongor Kiss who wrote nearly all of settings.ghci. I've just
adding a few little tweaks that make it so that the loaded ghci can be run from
within ghci. There were two things that needed to be changed:

1. The compiler code requires NoImplicitPrelude, but GHCi's modules are
incompatible with it. So, this adds the pragma to all GHCi modules that didn't
have it, and adds imports of Prelude.

2. In order to run GHC within itself, a call of 'initGCStatistics' needed to be
skipped. This uses CPP to skip it when -DGHC_LOADED_INTO_GHCI is set.


mgsloan:
Reviewers, feel free to bikeshed on my choices about naming and file locations.  Also, it may makes sense to not do this in the Makefile.  To me it seemed like a reasonably sensible spot to put a top level development command.

To run ghci configured with the settings (but no `:load`) do:

```
make run-ghci N=8
```

To run ghci and load `Main`, do:

```
make load-ghci N=8
```

Particularly not sure of the target name of `in-ghci`.  The idea is that it reads like "make in ghci", i.e. building within ghci.

mgsloan:
Oh, and one thing that should be resolved before merge is that on run of ghc within ghci via `:main`, it errors out with

```
/home/mgsloan/oss/haskell/ghc/inplace/lib/lib/settings: openFile: does not exist (No such file or directory)                         
```

To work around this I did the following:
```
$ cd inplace/lib
$ mkdir lib
$ cd lib
$ ln -s ../* .
```

alpmestan:
Probably because the actual executables are in `inplace/lib/bin`, and by default ghc looks for its `settings` file in a `lib` directory that should be right next to the bin directory that contains the ghc you're launching.

When calling ghc from the command line, this (the location of `lib`) is overridable with `-B`, which is exactly what the wrapper scripts from `inplace/bin` do. Maybe you'll want to find the programmatic way to do this?

mgsloan:
@alpmestan Thanks for the pointer!  Indeed it looks like that's what's going on.  Instead of requiring the user of `ghci` to provide `-B` manually, I've added a hack for topdir resolution that is only used when `-DGHC_LOADED_INTO_GHCI` is used.  Please let me know if you can think of a cleaner solution.

I've also pushed a commit which moves the invocation of `findTopDir` out of `initLlvmConfig` and `initSysTools`, because otherwise the output would look like:

```
λ :main                                                                                                                                                     
Note: Due to -DGHC_LOADED_INTO_GHCI, guessing that "/home/mgsloan/oss/haskell/ghc/inplace/lib" should be used as the topdir.                                
Note: Due to -DGHC_LOADED_INTO_GHCI, guessing that "/home/mgsloan/oss/haskell/ghc/inplace/lib" should be used as the topdir.                                
Note: Due to -DGHC_LOADED_INTO_GHCI, guessing that "/home/mgsloan/oss/haskell/ghc/inplace/lib" should be used as the topdir.
```

With that change, the message is no longer tripled.

alpmestan:
@mgsloan Great, glad you figured out a way to make this work as you want it to. The `GHC_LOADED_INTO_GHCI` bit and more generally the systools/`findTopDir`-related changes seem "important enough" to perhaps carry this patch over to a wider audience of GHC devs, on phabricator. Maybe someone will think of a better way to "override" the topdir resolution for your ghci use case. Would you mind opening a new differential there? (If you haven't used phabricator and its `arc` tool before, I can provide some guidance, it's very simple.)

mgsloan:
@alpmestan Good idea.  I opened the PR here since I am more familiar with github PRs, and initially this change was much smaller!  Thanks for the offer of guidance but reading the manual was sufficient.

Here's the

PR #158, opened 2018-07-01
Change function type fixity to (-1)

jhb563:
Addresses [#15235](https://ghc.haskell.org/trac/ghc/ticket/15235).

Tested by running `ghc/inplace/bin/ghc-stage2.exe --interactive`:

```
>> :i (->)
data (->) (a :: TYPE q) (b :: TYPE r) -- Defined . `GHC.Prim`
infixr -1 `(->)`
...
```

### DIFF
--- a/compiler/basicTypes/BasicTypes.hs
+++ b/compiler/basicTypes/BasicTypes.hs
@@ -409,7 +409,7 @@ defaultFixity = Fixity NoSourceText maxPrecedence InfixL
 negateFixity, funTyFixity :: Fixity
 -- Wired-in fixities
 negateFixity = Fixity NoSourceText 6 InfixL  -- Fixity of unary negate
-funTyFixity  = Fixity NoSourceText 0 InfixR  -- Fixity of '->'
+funTyFixity  = Fixity NoSourceText (-1) InfixR  -- Fixity of '->'
 
 {-
 Consider


bgamari:
@jhb563, thanks for the patch! Given that this is a rather trivial patch, I'll accept it via GitHub. That being said, it looks like CI failed due to a number of failing tests. Could you fix these? This should just be a matter of running `make -C testsuite accept`, reviewing the changes that result, and committing.

compiler/basicTypes/BasicTypes.hs
```diff
@@ -409,7 +409,7 @@ defaultFixity = Fixity NoSourceText maxPrecedence InfixL
 negateFixity, funTyFixity :: Fixity
 -- Wired-in fixities
 negateFixity = Fixity NoSourceText 6 InfixL  -- Fixity of unary negate
-funTyFixity  = Fixity NoSourceText 0 InfixR  -- Fixity of '->'
+funTyFixity  = Fixity NoSourceText (-1) InfixR  -- Fixity of '->'
```
sighingnow: Please add some comment here to explain why we need to change the fixity of `->`, and point to Trac #15235.

monoidal:
Documentation also needs updating; https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#infix-type-constructors-classes-and-type-variables states "Function arrow is infixr with fixity 0 (this might change; it’s not clear what it should be)."

RyanGlScott:
This has been superseded by https://git.haskell.org/ghc.git/commitdiff/251e3424a96986fca5164a2397783a1c066558fc, so this can be closed.

PR #165, opened 2018-07-14
Fix hash in haddock of ghc-prim.

msakai:
Some hash signs in documents in primops.txt.pp were not escaped properly. Those raw hash signs were kept in haddock and texts between them were interpreted as anchors by haddock.

Following is an example of problematic result.

https://hackage.haskell.org/package/ghc-prim-0.5.2.0/docs/GHC-Prim.html#v:copyByteArrayToAddr-35-
```
<p>Copy a range of the ByteArray<a id=" to the memory range starting at the Addr"></a>.
    The ByteArray<a id=" and the memory region at Addr"></a> must fully contain the
    specified ranges, but this is not checked. The Addr# must not point into the
    ByteArray<a id=" (e.g. if the ByteArray"></a> were pinned), but this is not checked
    either.</p>
```


bgamari:
Good catch. Thanks!

bgamari:
Merged in 28199a21b174ff1aff6a8acc8ec0253dca482205.

PR #168, opened 2018-07-21
docs: Fix wrong module name in hsig example

mboes:
In the module signatures section, two modules were defined, `Str` and
`A`, but `A` was importing `Text`, not `Str`.


bgamari:
Good catch; thanks!

bgamari:
Meged in b4302fd59b8ec127ad7011e6f6a4f969c78c486e.

PR #172, opened 2018-07-28
doc: Fix command for creating a shared library.

mboes:
The previous command did not work. The `-o` flag was missing.
Moreover, without `-dynamic`, the `-shared` flag will raise obscure
link errors because GHC will try to use static objects when creating
the shared library.


bgamari:
Thanks @mboes!

bgamari:
Merged in cb8efe708c57e5c91c592956de7e58d2c5ca5065.

PR #173, opened 2018-07-28
[docs] Add missed specialisations warnings to list of those not enabled by -Wall

MaxGabriel:
Enabling `-Weverything` does enable those warnings.

### DIFF
--- a/docs/users_guide/using-warnings.rst
+++ b/docs/users_guide/using-warnings.rst
@@ -82,6 +82,8 @@ The following flags are simple ways to select standard "packages" of warnings:
         * :ghc-flag:`-Widentities`
         * :ghc-flag:`-Wredundant-constraints`
         * :ghc-flag:`-Wpartial-fields`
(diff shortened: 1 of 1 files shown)


MaxGabriel:
Actually, I think the list also enables Safe Haskell warnings, which are disabled by `-Wno-unsafe`. While the specialisation warnings are in the list of warnings, `-Wunsafe` is missing entirely. Is that a mistake and should an entry be added, I presume?

bgamari:
Merged in 8edc4b4bac47d971b72b8ce62f847352cfe318e3.

> Actually, I think the list also enables Safe Haskell warnings, which are disabled by -Wno-unsafe. While the specialisation warnings are in the list of warnings, -Wunsafe is missing entirely. Is that a mistake and should an entry be added, I presume?

It sounds like this is the case. Could you open another pull request doing this?

MaxGabriel:
Sure, I opened #180 for that

PR #174, opened 2018-08-04
[base]:  Rewrite Functor documentation

gilligan:
This adds a (hopefully) more approachable, yet "correct enough" description of Functor along with some motivating examples.

This is my first PR for ghc. I hope I didn't get too much wrong and am looking forward to any feedback. Going forward i am hoping to invest some more effort into improving docs for base.

### DIFF
--- a/libraries/base/Data/Functor.hs
+++ b/libraries/base/Data/Functor.hs
@@ -11,8 +11,31 @@
 -- Stability   :  provisional
 -- Portability :  portable
 --
--- Functors: uniform action over a parameterized type, generalizing the
--- 'Data.List.map' function on lists.
+--
+-- A type @f@ is a Functor if it provides a function @fmap@ which, given any types @a@ and @b@,
+-- lets you apply any function of type @(a -> b)@ to turn an @f a@ into an @f b@, preserving the
+-- structure of @f@.
+--
+-- ==== __Examples__
+--
+--  >>> fmap show (Just 1)  --  (a   -> b)      -> f a       -> f b
+--  Just "1"                --  (Int -> String) -> Maybe Int -> Maybe String
+--
+--  >>> fmap show Nothing   --  (a   -> b)      -> f a       -> f b
+--  Nothing                 --  (Int -> String) -> Maybe Int -> Maybe String
+--
+--  >>> fmap show [1,2,3]   --  (a   -> b)      -> f a       -> f b
+--  ["1", "2", "3"]         --  (Int -> String) -> [Int]     -> [String]
+--
+--  >>> fmap show []        --  (a   -> b)      -> f a       -> f b
+--  []                      --  (Int -> String) -> [Int]     -> [String]
+--
+-- The 'fmap' function is also available as the infix operator '<$>':
+--
+--  >>> fmap show (Just 1) --  (Int -> String) -> Maybe Int -> Maybe String
+--  Just "1"
+--  >>> show <$> (Just 1)  --  (Int -> String) -> Maybe Int -> Maybe String
+--  Just "1"
 
 module Data.Functor
     (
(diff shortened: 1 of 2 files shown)


libraries/base/Data/Functor.hs
```diff
@@ -11,8 +11,29 @@
...
+--
+--  >>> fmap show [1,2,3]   -- fmap :: (a   -> b)      -> f a       -> f b
+--  ["1", "2", "3"]         -- fmap :: (Int -> String) -> [Int]     -> [String]
+--
+--  >>> fmap show []        -- fmap :: (a   -> b)      -> f a       -> f b
+--  []                      -- fmap :: (Int -> String) -> [Int]     -> [String]
+--
+-- The 'fmap' function is also available as the infix operator '<$>':
+--
+--  >>> show <$> (Just 1)  -- fmap :: (Int -> String) -> Maybe Int -> Maybe String
```
bgamari: I wonder whether the parens are helping or hurting here. I've seen newcomers have some trouble grasping Haskell's precedence scheme; I fear that adding extraneous parens, which making this example more explicit, may contribute to this confusion.

bgamari:
This looks great, @gilligan! What you have written here is correct and also a great improvement over the existing documentation.

I look forward to seeing more contributions along these lines. Thanks again!

libraries/base/Data/Functor.hs
gilligan: Hm, I only ever found the parens helpful myself, especially as a beginner. Also it’s consistent with the actual fmap implementation. I would really like to keep them unless this is a dealbreaker for you or anyone else.

gilligan:
@bgamari what am I to do about the ci failures? What can I do to move this PR forward? Should I ping anyone directly about a review?

libraries/base/Data/Functor.hs
argumatronic: I wonder if we could show a set of examples and talk about the precedence?

```
λ> show <$> Just 1
Just "1"
it :: Maybe String

λ> show <$> (Just 1)
Just "1"
it :: Maybe String

λ> fmap show Just 1
"Just 1"
it :: String

λ> fmap show (Just 1)
Just "1"
it :: Maybe String
```
I think that newcomers are probably surprised by the difference in behavior of the prefix vs infix versions of `fmap` as above. Maybe we could explain it and/or provide the examples instead of simply giving the `infixl` without comment? Precdence does take a lot of getting used to.

gilligan: Yes and no - while my PRs are all about making the docs as approachable and "user friendly" as possible i am unsure about this. If we elaborate on this we are not strictly talking about Functor anymore we are discussing how precedence works - and then this might as well be placed anywhere else in the documentation of any other module, or all of them, or none of them?

:thinking:

gilligan:
```
--  >>> fmap show (Just 1) --  (Int -> String) -> Maybe Int -> Maybe String
--  Just "1"
--  >>> show <$> (Just 1)  --  (Int -> String) -> Maybe Int -> Maybe String
--  Just "1"
```

not sure if this is perhaps some sort of compromise - illustrating that the two are equivalent.

argumatronic: Yeah, that's fair -- there probably isn't a good place for this, which may be why it remains confusing. The examples are, at any rate, specifically about the infix operator `<$>` and how it differs from the prefix `fmap`.

bgamari:
Don't worry about the CI failures; I cancelled the builds since your changes have a low probability of regression and if they do regress I will catch them when I merge (which I'll do now).

bgamari:
Merged with ce47a9c16f47511575925a6958d167241c4b6c89.

PR #176, opened 2018-08-08
base: Rewrite semigroup documentation

gilligan:
This is a rewrite of the documentation for `Semigroup`. I try explain the abstraction in an approachable yet hopefully sufficiently correct way and also provide several examples. Added is also a short paragraph for theoretical context which mentions the relation in which it exists together with `Monoid`.

The current documentation focuses completely on mathematics and gets into anecdotes - I tend to believe this is not beneficial for what is supposed to be user documentation.

### DIFF
--- a/libraries/base/Data/Semigroup.hs
+++ b/libraries/base/Data/Semigroup.hs
@@ -18,16 +18,53 @@
 -- Maintainer  :  libraries@haskell.org
 -- Stability   :  provisional
 -- Portability :  portable
+-- 
+-- A type @a@ is a 'Semigroup' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one. Where being 
+-- associative means that the following must always hold:
 --
--- In mathematics, a semigroup is an algebraic structure consisting of a
--- set together with an associative binary operation. A semigroup
--- generalizes a monoid in that there might not exist an identity
--- element. It also (originally) generalized a group (a monoid with all
--- inverses) to a type where every element did not have to have an inverse,
--- thus the name semigroup.
+-- >>> (a <> b) <> c == a <> (b <> c)
 --
--- The use of @(\<\>)@ in this module conflicts with an operator with the same
--- name that is being exported by Data.Monoid. However, this package
+-- ==== __Examples__
+--
+-- The 'Min' 'Semigroup' instance for 'Int' is defined to always pick the smaller
+-- number:
+-- >>> Min 1 <> Min 2 <> Min 3 <> Min 4 :: Min Int
+-- Min {getMin = 1}
+--
+-- If we need to combine multiple values we can use the 'sconcat' function
+-- to do so. We need to ensure however that we have at least one value to
+-- operate on, since otherwise our result would be undefined. It is for this
+-- reason that 'sconcat' uses "Data.List.NonEmpty.NonEmpty" - a list that
+-- can never be empty:
+--
+-- >>> (1 :| [])
+-- 1 :| []               -- equivalent to [1] but guaranteed to be non-empty
+-- >>> (1 :| [2, 3, 4])
+-- 1 :| [2,3,4]          -- equivalent to [1,2,3,4] but guaranteed to be non-empty
+--
+-- Equipped with this guaranteed to be non-empty data structure, we can combine
+-- values using 'sconcat' and a 'Semigroup' of our choosing. We can try the 'Min'
+-- and 'Max' instances of 'Int' which pick the smallest, or largest number
+-- respectively:
+--
+-- >>> sconcat (1 :| [2, 3, 4]) :: Min Int
+-- Min {getMin = 1}
+-- >>> sconcat (1 :| [2, 3, 4]) :: Max Int
+-- Max {getMax = 4}
+--
+-- String concatenation is another example of a 'Semigroup' instance:
+--
+-- >>> "foo" <> "bar"
+-- "foobar"
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Yet unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. The requirement for a neutral element prevents many types from being a full Monoid,
+-- like "Data.List.NonEmpty.NonEmpty".
+--
+-- Note that the use of @(\<\>)@ in this module conflicts with an operator with the same
+-- name that is being exported by "Data.Monoid". However, this package
 -- re-exports (most of) the contents of Data.Monoid, so to use semigroups
 -- and monoids in the same package just
 --


libraries/base/Data/Semigroup.hs
```diff
@@ -483,8 +496,6 @@ instance Applicative Option where
 
   Option Nothing  *>  _ = Option Nothing
   _               *>  b = b
-
--- | @since 4.9.0.0
```
parsonsmatt: Why was this removed? `@since` tags are fantastic for figuring out version bounds at a glance.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,20 +18,33 @@
...
--- In mathematics, a semigroup is an algebraic structure consisting of a
--- set together with an associative binary operation. A semigroup
--- generalizes a monoid in that there might not exist an identity
--- element. It also (originally) generalized a group (a monoid with all
--- inverses) to a type where every element did not have to have an inverse,
--- thus the name semigroup.
+-- Intuitively, this gives rise to an abstraction that allows for the
+-- combination of multiple values into a single one. The instance of 
+-- the 'Semigroup' determines the way in which the values are to be 
+-- combined.
```
parsonsmatt: This paragraph doesn't seem to add anything that the previous paragraph didn't already cover.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,20 +18,33 @@
...
+-- >>> First 1 <> First 2 <> First 3
+-- First {getFirst = 1}
 --
--- > import Data.Semigroup
+-- ==== __Theoretical Context__
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. This also lifts the restriction of 'Semigroup' to work only on non-empty
+-- types (as can be seen by the use of the NonEmpty list type instead of a normal list).
```
parsonsmatt: I'm unsure what the final sentence means. Could you elaborate?

libraries/base/Data/Semigroup.hs
```diff
@@ -18,20 +18,33 @@
...
--- thus the name semigroup.
+-- Intuitively, this gives rise to an abstraction that allows for the
+-- combination of multiple values into a single one. The instance of 
+-- the 'Semigroup' determines the way in which the values are to be 
+-- combined.
 --
--- The use of @(\<\>)@ in this module conflicts with an operator with the same
--- name that is being exported by Data.Monoid. However, this package
--- re-exports (most of) the contents of Data.Monoid, so to use semigroups
--- and monoids in the same package just
```
parsonsmatt: This paragraph seems like a useful thing to keep around.

argumatronic: Yeah I agree.

libraries/base/Data/Semigroup.hs
gilligan: maybe you are right - i was so stuck on trying to make it "approachable" but maybe this is just redundant. Unless anyone has any ideas on talking about the motivation / usecases...

libraries/base/Data/Semigroup.hs
gilligan: thanks, that was non-intentional.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,37 @@
...
 --
 -- > import Data.Semigroup
 --
+-- ==== __Theoretical Context__
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. The neutral element furthermore lifts the restriction of 'Semigroup' to work 
+-- only on non-empty types (as can be seen by the use of the NonEmpty list type instead 
+-- of a normal list).
```
parsonsmatt: OK, this wording is more clear, but now I think it is wrong :o Semigroup is not restricted to work only on non-empty types. Since Semigroup is a generaliation of Monoid, all Monoids are semigroups, but there are some semigroups which cannot provide a lawful `mempty` value and have a Monoid instance.

The requirement for a neutral element doesn't lift a restriction; it *incurs* a restriction on `Monoid`al types. I think I would word this as:

> The requirement for a neutral element forbids many types from being a full `Monoid`, like `NonEmpty`.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,37 @@
...
+-- >>> sconcat (1 :| [2, 3, 4]) :: Min Int
+-- Min {getMin = 1}
+-- >>> sconcat (1 :| [2, 3, 4]) :: Max Int
+-- Max {getMax = 4}
+-- >>> "foo" <> "bar"
+-- "foobar"
+-- >>> Last 1 <> Last 2 <> Last 3
+-- Last {getLast 3}
+-- >>> First 1 <> First 2 <> First 3
+-- First {getFirst = 1}
```
parsonsmatt: Examples <3

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
...
 -- re-exports (most of) the contents of Data.Monoid, so to use semigroups
 -- and monoids in the same package just
 --
 -- > import Data.Semigroup
 --
+-- ==== __Theoretical Context__
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. The requirement for a neutral element forbids many types from being a full Monoid,
```
gilligan: @parsonsmatt using your wording now, thanks for that

libraries/base/Data/Semigroup.hs
gilligan: fair, it's back in ;)

libraries/base/Data/Semigroup.hs
coot: Maybe it would be nice to add a word why `NonEmpty` appears in the class instance (if it's not too hard to understand)?
```
@NonEmpty a@ is a semigroup for each @a@ and this the free semigroup, i.e. there exists 
@
  foldMapSemigroup :: Semigroup s => (a -> s) -> NonEmpty a -> s
@
@sconcat@ is then just @foldMapSemigroup id@, moreover @foldMapSemigroup@ is analogous to 
@foldMap :: Monoid m => (a -> m) -> [a] -> m@.
```

Unfortunately there is no `foldMapSemigroup` in `Semigroup`, maybe it's worth adding it?
```
foldMapSemigroup :: Semigroup s => (a -> s) => NonEmpty a -> s
foldMapSemigroup f = sconcat . fmap f
```

coot: Anyway `sconcat` will be there and it might be good to know where it comes from, plus you learn a pattern that is repeating else where (and it is repeating itself all over Haskell ecosystem).

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
...
+-- 
+-- A type @a@ is a 'Semigroup' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one. 
 --
--- In mathematics, a semigroup is an algebraic structure consisting of a
--- set together with an associative binary operation. A semigroup
--- generalizes a monoid in that there might not exist an identity
--- element. It also (originally) generalized a group (a monoid with all
--- inverses) to a type where every element did not have to have an inverse,
--- thus the name semigroup.
```
llelf: Can we please not remove useful documentation?

gilligan: The fact that it generalizes the concept of a Monoid is mentioned below under "Theoretical Context". As for the rest: I am having a very hard time understanding how this is relevant or helpful at all.

How is the fact that it originally generalized a group, which is just a  monoid with all inverses (i actually don't understand what that even means) useful documentation? Who would miss this piece of information if it was not there and/or be less able to make use of Semigroup? Isn't this the kind of information that the avid user can look for at nlab or wikipedia?

if you and others really insist on including the information on how the name came to be then I suppose this can be added under "Theoretical Context", but please don't make me start documentation with "In mathematics" and explanations on the origins of the name. Shouldn't documentation focus first on the *purpose* and how it can be used and applied?

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
 -- Maintainer  :  libraries@haskell.org
 -- Stability   :  provisional
 -- Portability :  portable
+-- 
+-- A type @a@ is a 'Semigroup' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one.
```
bgamari: It would be good to define "associative" here. Perhaps something like,
```
Being associative, the function must obey the following law,
>>> (a <> b) <> c  ==  a <> (b <> c)
```

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
...
+-- "foobar"
+-- >>> Last 1 <> Last 2 <> Last 3
+-- Last {getLast 3}
+-- >>> First 1 <> First 2 <> First 3
+-- First {getFirst = 1}
 --
--- The use of @(\<\>)@ in this module conflicts with an operator with the same
+-- Note that the use of @(\<\>)@ in this module conflicts with an operator with the same
 -- name that is being exported by Data.Monoid. However, this package
 -- re-exports (most of) the contents of Data.Monoid, so to use semigroups
```
bgamari: Can you also fix this to be a proper module reference? That is, enclose `Data.Monoid` in double-quotes.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
...
 -- and monoids in the same package just
 --
 -- > import Data.Semigroup
 --
+-- ==== __Theoretical Context__
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. The requirement for a neutral element forbids many types from being a full Monoid,
+-- like NonEmpty.
```
bgamari: Perhaps make this a proper reference as well; e.g. `'Data.List.NonEmpty.NonEmpty'`.

libraries/base/Data/Semigroup.hs
```diff
@@ -18,21 +18,36 @@
...
+-- that lets you combine any two values of type @a@ into one. 
 --
--- In mathematics, a semigroup is an algebraic structure consisting of a
--- set together with an associative binary operation. A semigroup
--- generalizes a monoid in that there might not exist an identity
--- element. It also (originally) generalized a group (a monoid with all
--- inverses) to a type where every element did not have to have an inverse,
--- thus the name semigroup.
+-- ==== __Examples__
+-- >>> sconcat (1 :| [2, 3, 4]) :: Min Int
```
bgamari: As suggested in #177, it wouldn't hurt to first briefly introduce the structures being used here. Currently there is a bit of assumed familiarity that a new-comer may not possess.

gilligan:
@bgamari i reworked the examples a bit adding more explanations.Let me know what you think

libraries/base/Data/Semigroup.hs
```diff
@@ -18,16 +18,53 @@
...
+-- operate on, since otherwise our result would be undefined. It is for this
+-- reason that 'sconcat' uses "Data.List.NonEmpty.NonEmpty" - a list that
+-- can never be empty:
+--
+-- >>> (1 :| [])
+-- 1 :| []               -- equivalent to [1] but guaranteed to be non-empty
+-- >>> (1 :| [2, 3, 4])
+-- 1 :| [2,3,4]          -- equivalent to [1,2,3,4] but guaranteed to be non-empty
+--
+-- Equipped with this, guaranteed to be non-empty data structure, we can combine
```
alech: I think the first comma can and should be skipped

libraries/base/Data/Semigroup.hs
```diff
@@ -18,16 +18,53 @@
...
+-- values using 'sconcat' and a 'Semigroup' of our choosing. We can try the 'Min'
+-- and 'Max' instances of 'Int' which pick the smallest, or largest number
+-- respecitvely:
```
alech: respectively

libraries/base/Data/Semigroup.hs
```diff
@@ -18,16 +18,53 @@
...
+--
+-- >>> sconcat (1 :| [2, 3, 4]) :: Min Int
+-- Min {getMin = 1}
+-- >>> sconcat (1 :| [2, 3, 4]) :: Max Int
+-- Max {getMax = 4}
+--
+-- String concatenation is a nother example of a valid 'Semigroup' instance:
```
alech: another

alech: would skip the "valid", since why would you present an invalid Semigroup instance?

libraries/base/Data/Semigroup.hs
```diff
@@ -18,16 +18,53 @@
...
+--
+-- >>> "foo" <> "bar"
+-- "foobar"
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Yet unlike the 'Semigroup', the 'Monoid'
+-- requires the presence of a neutral element ('mempty') in addition to the associative
+-- operator. The requirement for a neutral element forbids many types from being a full Monoid,
```
alech: maybe "prevents" instead of "forbids"?

libraries/base/Data/Semigroup.hs
gilligan: fair enough, let's drop the valid.

libraries/base/Data/Semigroup.hs
gilligan: yeah, that sounds better.

gilligan:
@alech thanks, all concerns addressed (i hope)

gilligan:
@bgmari does this look mergable to you now? Anything you’d still like me to change?

libraries/base/Data/Semigroup.hs
```diff
@@ -18,16 +18,53 @@
...
+-- Min {getMin = 1}
+-- >>> sconcat (1 :| [2, 3, 4]) :: Max Int
+-- Max {getMax = 4}
+--
+-- String concatenation is another example of a 'Semigroup' instance:
+--
+-- >>> "foo" <> "bar"
+-- "foobar"
+--
+-- A 'Semigroup' is a generalization of a 'Monoid'. Yet unlike the 'Semigroup', the 'Monoid'
```
bgamari: Either say "Monoid" or "the Monoid class"

bgamari:
Merged.

libraries/base/Data/Semigroup.hs
gilligan: Sorry, seems like i totally forgot about this one!

PR #177, opened 2018-08-08
base: rewrite Monoid module docs

gilligan:
This updates the docs for Monoid analogous to the way i updated the description for Semigroup in #176

### DIFF
--- a/libraries/base/Data/Monoid.hs
+++ b/libraries/base/Data/Monoid.hs
@@ -16,8 +16,43 @@
 -- Stability   :  experimental
 -- Portability :  portable
 --
--- A class for monoids (types with an associative binary operation that
--- has an identity) with various general-purpose instances.
+-- A type @a@ is a 'Monoid' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one, and a neutral
+-- element (`mempty`) such that 
+--
+-- > a <> mempty == mempty <> a == a
+--
+-- A 'Monoid' is a 'Semigroup' with the added requirement of a neutral element.
+-- Thus any 'Monoid' is a 'Semigroup', but not the other way around.
+--
+-- ==== __Examples__
+--
+-- The 'Sum' monoid is defined by the numerical addition operator and `0` as neutral element:
+--
+-- >>> mempty :: Sum Int
+-- Sum 0
+-- >>> Sum 1 <> Sum 2 <> Sum 3 <> Sum 4 :: Sum Int
+-- Sum {getSum = 10}
+--
+-- We can combine multiple values in a list into a single value using the `mconcat` function. 
+-- Note that we have to specify the type here since 'Int' is a monoid under several different
+-- operations:
+--
+-- >>> mconcat [1,2,3,4] :: Sum Int
+-- Sum {getSum = 10}
+-- >>> mconcat [] :: Sum Int
+-- Sum {getSum = 0}
+--
+-- Another valid monoid instance of 'Int' is 'Product' It is defined by multiplication
+-- and `1` as neutral element:
+--
+-- >>> Product 1 <> Product 2 <> Product 3 <> Product 4 :: Product Int
+-- Product {getProduct = 24}
+-- >>> mconcat [1,2,3,4] :: Product Int
+-- Product {getProduct = 24}
+-- >>> mconcat [] :: Product Int
+-- Product {getProduct = 1}
+--
 --
 -----------------------------------------------------------------------------
 


Daniel-Diaz:
The module description explains `<>` and `mempty`, but then all the examples use `mconcat` instead, a function that has not been introduced yet at that point. For someone reading this for the first time, it can be a little confusing.

gilligan:
@Daniel-Diaz i see. Hm hm hm, ok let me try something.. hang on

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,29 @@
...
--- A class for monoids (types with an associative binary operation that
--- has an identity) with various general-purpose instances.
+-- A type @a@ is a 'Monoid' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one, and a neutral
+-- element (`mempty`) such that 
+--
+-- > a <> mempty == mempty a == a
+--
+-- ==== __Examples__
+-- >>> Sum 1 <> Sum 2 <> Sum 3 <> Sum 4 :: Sum Int
```
gilligan: @Daniel-Diaz how about now? I start of both scenarios (Sum and Product) with the invoking `<>` first, then use `mconcat`. I believe this should make it easier to deduce what is going on even without having previously seen `mconcat`.

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,29 @@
 -- Stability   :  experimental
 -- Portability :  portable
 --
--- A class for monoids (types with an associative binary operation that
--- has an identity) with various general-purpose instances.
+-- A type @a@ is a 'Monoid' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one, and a neutral
+-- element (`mempty`) such that 
+--
+-- > a <> mempty == mempty a == a
```
jez: I think you are missing a `<>`?

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,29 @@
...
--- A class for monoids (types with an associative binary operation that
--- has an identity) with various general-purpose instances.
+-- A type @a@ is a 'Monoid' if it provides an associative function ('<>')
+-- that lets you combine any two values of type @a@ into one, and a neutral
+-- element (`mempty`) such that 
+--
+-- > a <> mempty == mempty <> a == a
+--
+-- ==== __Examples__
+-- >>> Sum 1 <> Sum 2 <> Sum 3 <> Sum 4 :: Sum Int
```
bgamari: I wonder if we should add a sentence briefly introducing `Sum` so it is clear to a new-comer what this example means. For instance,

```
Take, for instance, the example of the 'Sum' monoid, defined by the numerical 
addition operator and the neutral element `0`:
>>> mempty :: Sum Int
Sum 0
>>> Sum 1 <> Sum 2 <> Sum 3 <> Sum 4 :: Sum Int
Sum {getSum = 10}
```
and so on.

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,29 @@
...
+-- Sum {getSum = 10}
+-- >>> mconcat [1,2,3,4] :: Sum Int
+-- Sum {getSum = 10}
+-- >>> mconcat [] :: Sum Int
+-- Sum {getSum = 0}
+-- >>> Product 1 <> Product 2 <> Product 3 <> Product 4 :: Product Int
```
bgamari: Similarly, we might also consider a brief sentence introducing `Product`.

libraries/base/Data/Monoid.hs
gilligan: done, PTAL

libraries/base/Data/Monoid.hs
gilligan: done, PTAL

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,42 @@
...
+-- The 'Sum' monoid is defined by the numerical addition operator and `0` as neutral element:
+--
+-- >>> mempty :: Sum Int
+-- Sum 0
+-- >>> Sum 1 <> Sum 2 <> Sum 3 <> Sum 4 :: Sum Int
+-- Sum {getSum = 10}
+--
+-- We can combine multiple values in a list into a single value using the `mconcat` function. 
+-- Note that we have to specify the type here since Int has more than one monoid instance:
+--
```
treeowl: Should be `'Int'`. I wouldn't say it *has" more than one "monoid instance". Perhaps "supports"?

treeowl: Better, perhaps: is a monoid under several different operations.

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,42 @@
...
+-- >>> mconcat [1,2,3,4] :: Sum Int
+-- Sum {getSum = 10}
+-- >>> mconcat [] :: Sum Int
+-- Sum {getSum = 0}
+--
+-- Another valid monoid instance of Int is 'Product' It is defined by multiplication
+-- and `1` as neutral element:
```
treeowl: `'Int'` or `@Int@`. Never just `Int`.

libraries/base/Data/Monoid.hs
```diff
@@ -16,8 +16,42 @@
...
+-- >>> Product 1 <> Product 2 <> Product 3 <> Product 4 :: Product Int
+-- Product {getProduct = 24}
+-- >>> mconcat [1,2,3,4] :: Product Int
+-- Product {getProduct = 24}
+-- >>> mconcat [] :: Product Int
+-- Product {getProduct = 1}
+--
+-- ==== __Theoretical Context__
+-- A 'Monoid' is a 'Semigroup' with the added requirement of a neutral element.
+-- Thus any 'Monoid' is a 'Semigroup', but not the other way around.
```
treeowl: This is not "theoretical context". It looks like mostly a restatement of what's up top. Remove it and integrate whatever you like into the beginning.

libraries/base/Data/Monoid.hs
gilligan: fixed

libraries/base/Data/Monoid.hs
gilligan: fair enough ;) I also just did the same for the semigroup PR..

libraries/base/Data/Monoid.hs
gilligan: That does indeed sound much better. Fixed!

gilligan:
@treeowl all problems fixed? Anything else you’d like to see changed before you’d be willing to ack the changes?

gilligan:
@treeowl gentle ping 😅

gilligan:
Any chance of getting this merged to master?

bgamari:
Merged. Thanks @gilligan!

PR #179, opened 2018-08-09
Make ghci work for stage1 and Hadrian

ndmitchell:
This allows you to use the ghc-in-ghci script in Make even if you haven't built stage2, and also with Hadrian.
Because of the way -I works, it will prefer Make/stage2.

CC @mgsloan (for review) and @snowleopard (for info)

### DIFF
--- a/utils/ghc-in-ghci/settings.ghci
+++ b/utils/ghc-in-ghci/settings.ghci
@@ -25,9 +25,6 @@
 :set -icompiler/vectorise
 :set -ighc
 :set -Icompiler
-:set -Icompiler/stage2
-:set -Icompiler/stage2/build
-:set -icompiler/stage2/build
 :set -Iincludes
 :set -Iincludes/dist-derivedconstants/header
 :set -package=ghc-boot-th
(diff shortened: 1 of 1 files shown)


snowleopard:
Awesome, thanks @ndmitchell!

mgsloan:
Cool, good stuff!

Initially I was a bit concerned that the user could get some really strange results if some files are missing from the header include dirs.  However, it looks like the only `*.h` files are `build/autogen/cabal_macros.h` and `ghc_boot_platform.h`.  So, this seems quite fine to me!

If there are consistency issues, then we could consider splitting each one of these out into a separate .script file.  Then, `utils/ghc-in-ghci/run.sh` could have logic to choose the appropriate script.

bgamari:
Hmm, this does seem a bit fragile but I suppose we can see how it goes.

mgsloan:
Hmm, yeah, actually how does this work when you have only built stage 1?  I assume it is using the host's ghc?  I'm actually a bit surprised that works very easily, but it would be great if it does, because it would be nice to offer a fast path for getting started with GHC development.

ndmitchell:
If you only build stage1 it works just fine to check the thing compiles, but if you try and use the resulting stage1 then you get errors about the wrong .hi file format - so it's great for feedback.

I'm keen to keep the simplicity of the existing system - it "just works". Every time you have generated files, or have to get a user to "pick" which flavour of ghcid they'd like, it really reduces the ease of use.

ndmitchell:
For stage1 I compiled it using a stage1 binary I had already built.

monoidal:
This was pushed some time ago in 8154faf5a7236f.

PR #180, opened 2018-08-09
Add warnings to list that -Wall does not enable

MaxGabriel:


MaxGabriel:
The Safe Haskell warnings aren't documented on the warnings page at all. Instead they're here: https://downloads.haskell.org/~ghc/master/users-guide/safe_haskell.html#ghc-flag--Wunsafe

Should they be added to the warnings page?

bgamari:
@MaxGabriel, it would make sense to mention them but I believe Sphinx throws a warning if a `:ghc-flag:` is defined more than once. Perhaps just add a reference?

MaxGabriel:
Something like “in addition to these warnings, see the Safe Haskell guide for additional Safe Haskell related warnings”?

bgamari:
Yes, that sounds

DamienCassou:
What is the status of this PR? I'm still confused about what each option enable or doesn't enable.

monoidal:
Since 46120bb637452, we list warnings that are enabled by `-Wall`, rather than listing those which are not. Closing.

PR #183, opened 2018-08-13
No implicit prelude in ghc-boot-th, ghc-boot and and ghci

shayne-fletcher-da:
Under the mentorship of @ndmitchell, add `import Prelude` where necessary to files in `compiler`, `ghc-boot-th`, `ghc-boot` and `ghci`, update cabal files with `default-extensions: NoImplicitPrelude`.


ndmitchell:
Closed in favour of #184

PR #187, opened 2018-08-21
Fix Typos/Mistakes in User Guide

staffehn:
(About the `f a b = R { .. }` line and nearby changes:) For some reason in 5d89565b043eaff9028205b79363ef0d0c17ff17 the argument `a` was removed; also the ` qualified` was dropped in that commit, making the `M.` below redundant (which I thus removed).

To double check I actually compiled the example, taking the missing-fields warnings as an indicator for what the expansion must have been. The compiler then reminded me of the fact that the constructor `R` has to be imported, too, an error even older that that commit linked above.

I double checked the rest of the user guide but there is no other example with an import where the constructor might have been forgotten like this.

### DIFF
--- a/docs/users_guide/glasgow_exts.rst
+++ b/docs/users_guide/glasgow_exts.rst
@@ -3545,10 +3545,10 @@ More details:
        module M where
          data R = R { a,b,c :: Int }
        module X where
-         import M( R(a,c) )
-         f b = R { .. }
+         import M( R(R,a,c) )
+         f a b = R { .. }
 
-   The ``R{..}`` expands to ``R{M.a=a}``, omitting ``b`` since the
+   The ``R{..}`` expands to ``R{a=a}``, omitting ``b`` since the
    record field is not in scope, and omitting ``c`` since the variable
    ``c`` is not in scope (apart from the binding of the record selector
    ``c``, of course).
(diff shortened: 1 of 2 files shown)


staffehn:
Added another fixing a different (and simpler) typo in the same file.

staffehn:
Added another commit to the same pull request, since it too is about mistakes in the user guide.

Here I noticed inconsistencies between the information if a particular warning is enabled by default that’s provided at the individual warning’s description vs. which warnings are listed in the compiled list of the whole set of all standard warnings at the top of the user-guide page.

I double-checked `-Wduplicate-constraints` indeed being not a standard warning in my ghc 8.4.3, since it doesn’t say what’s the case at the warnings’s description. I also double checked the four entries I added to the list, despite the fact that they are already stating to be a standard warning anyways. In case anyone reads that commit message, I didn’t actually find any missing warnings I didn’t double check; I probably got confused there ,and had no easy way to double check the diff of that list before creating the commit on GitHub’s editor.

I also went through the hole file, checking every occurrence of the phrase "by default" to make sure, no more warning, documented to be standard, is missing from the list, and also went through the list to check no listed warning is documented not to be standard.

staffehn:
Reminder: Is this getting merged at any time?

bgamari:
Yes. Merging now.

bgamari:
Merged in 4e3f6a0767b57dfa54934eb4fdb27f54db7016ee.

PR #190, opened 2018-08-30
Users guide: Mention that EmptyDataDecls is on by default

chris-martin:
The current phrasing seems to imply that `EmptyDataDecls` is required to define a void type. But in Haskell 2010, that isn't true. I suspect this writing is just outdated from pre-2010.


osa1:
Thanks! Merged as 4edc6d64d1bc1898c0974cf26c5713a3b2724a0b.

PR #192, opened 2018-09-03
Allow uppercase OverloadedLabels (#11671)

howtonotwin:
Implements [this feature request](https://ghc.haskell.org/trac/ghc/ticket/11671).


osa1:
Thanks!

This is a user-facing change and needs to go through the [GHC proposals](https://github.com/ghc-proposals/ghc-proposals/) process. Very briefly looked at the ticket, it seems like this is a backwards compatible change and there are use cases and requests from other users, so I suspect it'll be an easy accept. Would you mind writing a proposal for this change and submit a PR to the ghc-proposals repo?

PR #194, opened 2018-09-07
Avoid creating unevaluated Int thunks when iterating in GHC.Foreign

ndmitchell:
Currently the code builds up progressively larger `+1` calls, as the `Int` is only evaluated in debug mode. If instead we use a bang pattern then it uses constant memory.

I didn't actually measure performance, but before the code was the classic `foldl` pattern, and now it's making it `foldl'`. I stumbled across this code when debugging a space leak (this wasn't the code triggering the symptom I saw, but it is still a space leak).


osa1:
Thanks! Normally PRs are for documentation changes and we ask people to submit code changes as a Phabricator diff but I guess this is simple enough that can merge directly.

Can you please edit the commit line? At most 100 characters are allowed. (othewise GHC git server rejects it)

ndmitchell:
Repushed with a shorter title.

I've never used Phab before so the effort to learn is not justified (to me) by such a small patch...

osa1:
Merged as 510c5f4f22aca29a9c36fd993ac79e9077b28173.

PR #198, opened 2018-09-23
Fix test for GCC support for atomics in Autotools

sideeffffect:
that check would always fail, even if support for atomic operations was available

```c
conftest.c: In function ‘test’:
conftest.c:1:27: error: size mismatch in argument 2 of ‘__atomic_load’
 int test(int *x) { int y; __atomic_load(&x, &y, __ATOMIC_SEQ_CST); return x; }
                           ^~~~~~~~~~~~~
conftest.c:1:75: warning: returning ‘int *’ from a function with return type ‘int’ makes integer from pointer without a cast [-Wint-conversion]
 int test(int *x) { int y; __atomic_load(&x, &y, __ATOMIC_SEQ_CST); return x; }
```

I noticed that this was fixed in https://github.com/ghc/ghc/commit/ce3897ffd6e7c8b8f36b8e920168bac8c7f836ae but was not preserved by a reverting commit https://github.com/ghc/ghc/commit/d7fa8695324d6e0c3ea77228f9de93d529afc23e

/cc @bgamari @angerman @iliastsi @monoidal

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -1303,7 +1303,7 @@ AC_DEFUN([FP_GCC_SUPPORTS__ATOMICS],
 [
    AC_REQUIRE([AC_PROG_CC])
    AC_MSG_CHECKING([whether GCC supports __atomic_ builtins])
-   echo 'int test(int *x) { int y; __atomic_load(&x, &y, __ATOMIC_SEQ_CST); return x; }' > conftest.c
+   echo 'int test(int *x) { int y; __atomic_load(x, &y, __ATOMIC_SEQ_CST); return y; }' > conftest.c
    if $CC -c conftest.c > /dev/null 2>&1; then
        CONF_GCC_SUPPORTS__ATOMICS=YES
        AC_MSG_RESULT([yes])


iliastsi:
Hi all,

I just saw d7fa869 which removes `-latomic` and I would like to notice that this library is needed to compile GHC on mips and mipsel. Here are the build logs for [mips](https://buildd.debian.org/status/fetch.php?pkg=ghc&arch=mips&ver=8.4.3-6&stamp=1537046321&raw=0) and [mipsel](https://buildd.debian.org/status/fetch.php?pkg=ghc&arch=mipsel&ver=8.4.3-6&stamp=1537179102&raw=0) where compilation of ghc-8.4.3 (which does not contain ec9aacf) fails with
```
 /usr/bin/ld: /<<PKGBUILDDIR>>/libraries/ghc-prim/dist-install/build/libHSghc-prim-0.5.2.0_p.a(atomic.p_o): in function `hs_atomicread64':
atomic.c:(.text.hs_atomicread64+0x20): undefined reference to `__sync_add_and_fetch_8'
/usr/bin/ld: atomic.c:(.text.hs_atomicread64+0x24): undefined reference to `__sync_add_and_fetch_8'
/usr/bin/ld: /<<PKGBUILDDIR>>/libraries/ghc-prim/dist-install/build/libHSghc-prim-0.5.2.0_p.a(atomic.p_o): in function `hs_atomicwrite64':
atomic.c:(.text.hs_atomicwrite64+0x44): undefined reference to `__sync_bool_compare_and_swap_8'
/usr/bin/ld: atomic.c:(.text.hs_atomicwrite64+0x48): undefined reference to `__sync_bool_compare_and_swap_8'
collect2: error: ld returned 1 exit status
```

@bgamari Could you please re-introduce -latomic for OSes on which libatomic exists?

bgamari:
It would be great if someone could offer a PR introducing `-latomic` again. However, we'll need to make sure we test it a bit more before merging this time.

sideeffffect:
@bgamari OK, but this PR is only about fixing something which is essentially a typo
is *this* PR as it is acceptable to the GHC developers?

monoidal:
Merged, thank you!

PR #201, opened 2018-10-02
Fix typo in documentation

janpath:
Pretty sure this is right, but better take a look when you merge this.


bgamari:
Yes, I suppose this true. Thanks!

PR #204, opened 2018-10-16
Ignore {-# HLINT #-} pragmas, as used by HLint

ndmitchell:
The newest (as yet unreleased) version of HLint can interpret `{-# HLINT ... #-}` to ignore hints, as per https://github.com/ndmitchell/hlint#ignoring-hints. This change ensures GHC doesn't warn that the pragma is unknown.

As suggested/requested by @simonmar and @angerman

### DIFF
--- a/compiler/parser/Lexer.x
+++ b/compiler/parser/Lexer.x
@@ -2937,7 +2937,7 @@ ignoredPrags = Map.fromList (map ignored pragmas)
                      impls = ["hugs", "nhc98", "jhc", "yhc", "catch", "derive"]
                      options_pragmas = map ("options_" ++) impls
                      -- CFILES is a hugs-only thing.
-                     pragmas = options_pragmas ++ ["cfiles", "contract"]
+                     pragmas = options_pragmas ++ ["cfiles", "contract", "hlint"]
 
 oneWordPrags = Map.fromList [
      ("rules", rulePrag),


bgamari:
Adding a case to GHC's lexer every time a tool wants a piece of syntax seems like an unsustainable approach. Is this really a precedent that we want to be setting? 

Isn't there non-reserved syntax that HLint could take? For instance `{-! HLINT ... -}`.

ndmitchell:
If HLint takes `{-!` then at some point in the future someone in GHC world will decide that people often misspell `{-!` and we'll have the problem all over again. The entire purpose of pragmas was to add exactly this kind of notes, so I think it's well within the spirit of Haskell. My hope is the number of additions of this scope will be very small over time. There is already the precedence that CATCH, YHC and DERIVE are all whitelisted, so I'd be happy to trade adding HLINT for removing one of those 😄, thus keep the list the same size.

treeowl:
I generally agree with @bgamari here, I think. Pragmas are the exclusive domain of the compiler. Preprocessors are welcome to use them too, but should strip out their own. It could be useful to have a single "official" convention for communicating with tools like HLint, Liquid Haskell, and Haddock that inspect Haskell source for other purposes. While we're at it, it might make sense to give a standard format for preprocessor pragmas to avoid pragma name collisions when multiple preprocessors are used, but that's orthogonal.

bgamari:
> It could be useful to have a single "official" convention for communicating with tools like HLint, Liquid Haskell, and Haddock that inspect Haskell source for other purposes. 

Yes, I agree with this. 

> There is already the precedence that CATCH, YHC and DERIVE are all whitelisted, so I'd be happy to trade adding HLINT for removing one of those 

The difference is that these are all pragmas of other compilers. Note that the Haskell Report uses the particular phrase "compiler pragma" when discussing the `{-# ... #-}` syntax. I think it would be quite reasonable to make the distinction between compiler pragmas and tooling pragmas, more generally. Liquid Haskell is a good example here: they chose the syntax `{-@ LIQUID ... @-}`. I think it would make a lot of sense to standardize this syntax of `{-@ $TOOL_NAME ... @-}` in the Report, in much the same way compiler pragmas are standardized.

int-index:
I am uneasy about making a special case for HLint here, but I don't find the proposed distinction between tooling/compiler pragmas useful either.

What about introducing `-fno-warn-pragma=XXX`? People who use HLint will add
`-fno-warn-pragma=HLINT` to their build configuration.

bgamari:
@ndmitchell, I've started a discussion on ghc-devs for this matter.

@int-index, a warning flag is an interesting way to deal with the issue. On the other hand, it's not great from an ergonomic perspective; afterall, this would mean that all users of HLint (and any other tool requiring special pragmas) include this flag in their build configuration.

I think it makes a lot of sense to have a standard way for third-parties to attach string-y information to Haskell source constructs. Perhaps instead of compiler vs. tool pragmas another way to think about the distinction is:

 * `{-# ... #-}` is a flat namespace of special directives common enough across all programs to warrant special treatment (e.g. `INLINE`)
 * `{-@ $TOOL_NAME ... @-}` is an extensible namespace for everything else

Of course, whether all compiler directives are really common enough to warrant this treatment is debatable (I would guess there are probably far more `HLINT` pragmas than `SPECIALIZABLE` pragmas on Hackage). But nevertheless, this is where we are today.

ndmitchell:
HLint 2.1.11 now supports `{- HLINT ignore -}` and `{-# HLINT ignore #-}`. I'd really prefer to direct people towards the second of those, but only if GHC accepts the change. Otherwise I'm going to go all the syntax highlighting tools for Haskell and make them highlight `{- HLINT -}` differently (which makes me a bit sad).

PR #205, opened 2018-10-19
Improve documentation for warning options

ntc2:
My main goal was to make it easy to discover how to reverse `-Werror`, since `Wno-error` doesn't work, and the fact that `-Wwarn` negates `-Werror` was only mentioned in the `-Wwarn` docs before.

Other changes:
- explain at the outset that some options control individual warnings while others control warning families.
- explain at the outset that `-Wno-<wflag>` can be used to reverse `-W<wflag>`. This is no mentioned in
  two places, but I don't think that's a bad thing.

### DIFF
--- a/docs/users_guide/using-warnings.rst
+++ b/docs/users_guide/using-warnings.rst
@@ -8,8 +8,13 @@ Warnings and sanity-checking
    single: warnings
 
 GHC has a number of options that select which types of non-fatal error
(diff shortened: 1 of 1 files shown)


monoidal:
This was pushed in b8e30e40b08db3dd50e7a972a4969c8f8f784eaa.

PR #209, opened 2018-10-25
Fix sample code of -fprint-explicit-kinds, plus sample when disabling PolyKinds

igrep:
- Although the sample is for `-fprint-explicit-kinds`, the sample code uses `-fprint-explicit-foralls` (but perhaps the output used to be correct one of `-fprint-explicit-kinds`).
- Update the output with the recent version of GHC (ver. 8.4.3. I guess it doesn't change in GHC 8.6...)
- Add more samples to clarify the difference of kinds, which tells effect of `-fprint-explicit-kinds` more clearly.

### DIFF
--- a/docs/users_guide/using.rst
+++ b/docs/users_guide/using.rst
@@ -781,13 +781,16 @@ messages and in GHCi:
 
     .. code-block:: none
 
-        ghci> :set -XPolyKinds
-        ghci> data T a = MkT
-        ghci> :t MkT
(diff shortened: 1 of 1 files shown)


bgamari:
Looks reasonable to me. Thanks!

bgamari:
Merged in 44a1d1f6612e1e9ef07626e1d2536bb6ba55dbd2.

PR #212, opened 2018-10-29
Correctly detect GIT in a subtree

hsyl20:
When we use a git subtree, .git is a file, not a directory.
The script was already fixed for the commit ID but not for its date.


monoidal:
Pushed, thank you!

PR #216, opened 2018-11-07
Merge master, part 8a

monoidal:


mpickering:
Wrong branch I think. Looks like you're having fun though.

PR #220, opened 2018-11-13
Another branch

DanielRrr:


compiler/hsSyn/HsTypes.hs
```diff
@@ -466,11 +466,15 @@ instance OutputableBndr HsIPName where
 --------------------------------------------------
 
 -- | Haskell Type Variable Binder
+data HasBraces = Yes | No
```
int-index: Since constructors are not namespaced, I think it would be more appropriate to use names like this:

```
data TvbBraces = TvbHasBraces | TvbNoBraces
```

compiler/hsSyn/HsUtils.hs
```diff
@@ -371,14 +371,15 @@ mkHsStringPrimLit fs
 
 -------------
 userHsLTyVarBndrs :: SrcSpan -> [Located (IdP (GhcPass p))]
+                  -> [HasBraces]
                   -> [LHsTyVarBndr (GhcPass p)]
 -- Caller sets location
-userHsLTyVarBndrs loc bndrs = [ L loc (UserTyVar noExt v) | v <- bndrs ]
+userHsLTyVarBndrs loc bndrs braces = [ L loc (UserTyVar noExt v brace) | v <- bndrs, brace <- braces ]
```
int-index: This doesn't look right. Did you meant to `zip` these lists instead of making a nested loop (each-with-each)?

compiler/hsSyn/HsUtils.hs
```diff
@@ -371,14 +371,15 @@ mkHsStringPrimLit fs
...
 
-userHsTyVarBndrs :: SrcSpan -> [IdP (GhcPass p)] -> [LHsTyVarBndr (GhcPass p)]
+userHsTyVarBndrs :: SrcSpan -> [IdP (GhcPass p)] -> [HasBraces] -> [LHsTyVarBndr (GhcPass p)]
 -- Caller sets location
-userHsTyVarBndrs loc bndrs = [ L loc (UserTyVar noExt (L loc v))
-                             | v <- bndrs ]
+userHsTyVarBndrs loc bndrs braces = [ L loc (UserTyVar noExt (L loc v) brace)
+                             | v <- bndrs, brace <- braces]
```
int-index: Same thing, why would you go through all combinations?

PR #228, opened 2018-11-23
circleci: install newer version of alex

monoidal:
Currently, CircleCI is broken because of

```configure: error: Alex version 3.1.7 or later is required to compile GHC.```


monoidal:
I'm not sure, but at least this fixes the configure error (and uncovers test failures).

bgamari:
I believe this should be sorted now.

PR #230, opened 2018-11-25
Fixed plugin example which didn't compile

pythonissam:
Using GHC 8.6.2, example code doesn't compile. Then, regarding [this issue](https://ghc.haskell.org/trac/ghc/wiki/Plugins/TypeChecker#FAQ), we need to add `OPTIONS_GHC`.

### DIFF
--- a/docs/users_guide/extending_ghc.rst
+++ b/docs/users_guide/extending_ghc.rst
@@ -739,12 +739,13 @@ displayed.
     import HsDoc
 
     plugin :: Plugin
-    plugin = defaultPlugin { parsedResultAction = parsedPlugin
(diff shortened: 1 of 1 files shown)


docs/users_guide/extending_ghc.rst
```diff
@@ -783,12 +782,13 @@ When you compile a simple module that contains Template Haskell splice
 
 ::
 
+    {-# OPTIONS_GHC -fplugin SourcePlugin #-}
```
mpickering: This part is implied below but I suppose it's fine to have in the file as well.

mpickering:
Thanks, looks good.

PR #231, opened 2018-11-26
Hadrian: Print all testsuite output to stdout

snowleopard:
See https://ghc.haskell.org/trac/ghc/ticket/15951

@bgamari Sorry, I'm away from my laptop and a short GitHub PR is the best I can do!


bgamari:
Thanks @snowleopard!

alpmestan:
Landed as https://github.com/ghc/ghc/commit/93e86d6103757b43017535c92bc6970e9e2315a5.

PR #232, opened 2018-12-05
Add some complexities to Data.List documentation (#15003)

supersven:
I guess, the title tells it all :smile: 

I hope it's okay to submit my work on #15003 in small chunks. This way is very convenient for me.

PR #234, opened 2018-12-08
Data.list complexities 2

supersven:
Follow-up to #232

These are commits, that aren't merged, yet (i.e. that were not part of #232).

I hope this workflow is fine for you. Appending to a pull request with already merged commits looked like it could lead to confusion ...

---

Goal: **Add some complexities to Data.List documentation (#15003)**


bgamari:
Yes, this workflow is fine. Thanks!

The patch looks good.

PR #235, opened 2018-12-09
Fix minor correctness issue

ggreif:
The alias is of type i8, so its global variable name
should have type i8*. Anyway we should never deal
with pointers to (i8*)!


bgamari:
Great catch!

PR #238, opened 2018-12-15
Hadrian: Don't suggest trying to build base with stage0

mpickering:
You can't build base with stage0 (at least easily by following these changes) so I removed this confusing example and left one which should work.


mpickering:
As a more general point, the use of `defaultPackages` here seems to be a bit fragile as it relies on the fact that we are overriding fields of `defaultFlavour` which is only by coincidence. It would be more robust to pass a `packages` function to `modifiedPackages` to modify.

PR #241, opened 2018-12-18
add error codes to the Show instance of IOException

andrewthad:
Resolves https://ghc.haskell.org/trac/ghc/ticket/16048

### DIFF
--- a/libraries/base/GHC/IO/Exception.hs
+++ b/libraries/base/GHC/IO/Exception.hs
@@ -413,9 +413,10 @@ userError str   =  IOError Nothing UserError "" str Nothing Nothing
 -- ---------------------------------------------------------------------------
 -- Showing IOErrors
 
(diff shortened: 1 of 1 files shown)


andrewthad:
Worth noting that I didn't attempt to compile this

andrewthad:
This has been [compiled and tested](https://ghc.haskell.org/trac/ghc/ticket/16048#comment:4), and the new behavior is provided in the trac ticket.

treeowl:
This needs a test. Also, the CI is failing; is that unrelated?

andrewthad:
Where should I put the test for this? I don't see a directory in the test suite that it naturally fits into.

treeowl:
@andrewthad, I'd put the test under `libraries/base/tests`, which is admittedly a bit of a mess.

PR #246, opened 2018-12-27
./boot: silence stdout in "git config" call.

bairyn:
This script checks if the git configuration for URL rewriting is present, if
the remote is set to GitHub.  It checks the return code of "git config
--get-regexp '^url.*github.com/.*/packages-.insteadOf'".  But it didn't silence
the output of this command, so it was being sent to standard output, and
probably ultimately to your pager.

Previously, running ./boot on my system without URL rewrite git configuration
started the "less" pager with a buffer containing the output I had to close
with 'q', and it wasn't immediately clear that the buffer wasn't related to my
interaction with the script.

(Unlike applying "git --no-pager config", this commit also silences the output
of "git config --get-regexp '^url.*github.com/.*/packages-.insteadOf'" once the
GitHub URL rewrite configuration is applied.)

### DIFF
--- a/boot
+++ b/boot
@@ -133,6 +133,7 @@ def boot_pkgs():
                                 dir = dir_)))
 
                 makefile = os.path.join(package, 'GNUmakefile')
+                print('Creating %s' % makefile)
                 with open(makefile, 'w') as f:
                     f.write(dedent(
                         """\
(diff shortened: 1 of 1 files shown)


monoidal:
There are three changes here.

* Submodule rewriting: I removed the entire function `check_for_url_rewrites` in 9be82389a70. Submodules now use absolute paths and this rewriting is no longer neccessary.

* Adding `--recursive` to `git submodule init`. As far as I know, we do not have recursive submodules: `git submodule foreach` and `git submodule foreach --recursive` give the same output. Maybe we had them in the past, I'm not sure. It's more principled to use `--recursive`, though I would leave this as-is.

* A print statement for `GNUmakefile`. This makes sense - I'll submit this change to gitlab.

bgamari:
The code which this affects no longer exists.

PR #249, opened 2019-02-08
Write friendlier documentation for Integer

chris-martin:
Since this is one of the first types that we expect users of Haskell to encounter, having reasonable documentation for it is important. This comment ends up in the documentation for `Prelude`, so it ought to explain what an `Integer` is and what it's for, and it should not include references to constructors which are not exported from `Prelude`.

The most interesting thing I could think to point out about `Integer` from a user's perspective is how its infinite nature contrasts with `Int`.

### DIFF
--- a/libraries/integer-gmp/src/GHC/Integer/Type.hs
+++ b/libraries/integer-gmp/src/GHC/Integer/Type.hs
@@ -136,7 +136,11 @@ instance Eq BigNat where
 instance Ord BigNat where
     compare = compareBigNat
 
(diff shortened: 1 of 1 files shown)


rockbmb:
@chris-martin I opened a merge request to GHC in GitLab [here](https://gitlab.haskell.org/ghc/ghc/merge_requests/308), I don't mind closing it in favor of this.

rockbmb:
That MR I wrote has been merged, you might want to update this.

bgamari:
Merged with 3364def0a28d8b888f558cb9fdab379685636854. Unfortunately it looks like authorship was lost; my apologies to @chris-martin.

PR #252, opened 2019-05-03
support small arrays and CONSTR_NOCAF in ghc-heap

dten:
ghc-heap currently gets stuck at certain closure types that seem to be missing from the code

please note first time attempting to change something in ghc 😅


osa1:
Thanks! This should be submitted as a MR on Gitlab: https://gitlab.haskell.org/ghc/ghc

PR #253, opened 2019-05-21
Don't lose parantheses in show SomeAsyncException

ntc2:
I'd also like to change the string-based show instances in this file so that they wrap the strings with spaces in parens when precedence is 11 or higher, but I worry that change might be controversial. But hopefully this simpler change is not controversial, since we just preserves any existing parens in the wrapped exception.


mpickering:
Please can you open this MR on gitlab?

ntc2:
@mpickering done: https://gitlab.haskell.org/ghc/ghc/merge_requests/1001. Sorry, I hadn't realized that trivial PRs were no longer accepted on GitHub.

PR #254, opened 2019-05-21
Improve ThreadId Show instance

ntc2:
By making it include parens when a derived instance would. For example, this changes the (hypothetical) code `show (Just (ThreadId 3))` to produce `"Just (ThreadId 3)"` instead of the current `"Just ThreadId 3"`.


ntc2:
Moved to GitLab: https://gitlab.haskell.org/ghc/ghc/merge_requests/1002

PR #255, opened 2019-05-31
Fix documentation for -fomit-yields

23Skidoo:
When `-fomit-yields` is on (default), additional yield points are not added.

The error was introduced in 2c23fff2e03e77187dc4d01f325f5f43a0e7cad2, which was a pure documentation patch and didn't change the default.

/cc @bgamari


nh2:
This seems like an easy change.

bgamari:
A similar change has been made upstream.

PR #259, opened 2020-06-25
Mention flags that are not enabled by -Wall (#18372)

vtols:
* Mention missing flags that are not actually enabled by -Wall (docs/users_guide/using-warnings.rst)
* Additionally remove -Wmissing-monadfail-instances from the list of flags enabled by -Wcompat, as it is not the case since 8.8


vtols:
Duplicates https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3623

PR #260, opened 2020-07-03
Document interaction between unsafe FFI and GC

alexanderkjeldaas:
In the multi-threaded RTS this can lead to hard to debug
performance issues.


alexanderkjeldaas:
I am trying to add some more guidance to the docs in this PR.  The multi-threaded RTS is the most relevant one IMO going forwards, but there is less focus on this mode than the single threaded IMO.

Particularly scaling and performance issues related to the FFI are important to focus on.

I understand that some of this is specific to a stop-the-world GC strategy, so if there are ways to avoid some of these issues by using a different type of GC that would be good extra information to add.

If the current GC has a name, maybe specifying that certain behavior is conditioned on running the "foo-GC" would be good to add.

PR #261, opened 2021-01-04
Do not mmap with PROC_EXEC|PROC_WRITE on AArch64 Darwin

hjelmn:
Attempting to map memory that is both executable and writable will always
fail on Apple Silicon hosts. This behavior is similar to iOS. This PR
fixes the issue by changing preloadObjectFile to map read/write not
read/write/exec on this platform.

Signed-off-by: Nathan Hjelm <hjelmn@cs.unm.edu>

### DIFF
--- a/rts/Linker.c
+++ b/rts/Linker.c
@@ -1523,8 +1523,8 @@ preloadObjectFile (pathchar *path)
       return NULL;
    }
 
-   /* iOS does not permit to mmap with r+w+x, however while the comment for
-    * this function says this is not the final resting place, for some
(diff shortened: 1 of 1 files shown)


bgamari:
@hjelmn can you move this over to https://gitlab.haskell.org/ghc/ghc?

hjelmn:
Yes. I will do that. I have another couple of fixes I am working on in the linker code. I have a version of 8.10.3 that passes most of the testsuite on Apple Silicon. I am forward-porting the patches to master.

PR #262, opened 2021-01-06
Fix calls to varargs C function fcntl

nrdmn:
The ccall calling convention doesn't support varargs functions, so
switch to capi instead. See
https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/ffi.html#varargs-not-supported-by-ccall-calling-convention


bgamari:
This looks good. I have moved it to [GitLab](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4755) for merging.

PR #265, opened 2021-10-10
Minor changes

snxx-lppxx:


rts/Task.h
```diff
@@ -79,6 +79,8 @@
           ownership of the Task and a Capability.
 */
 
+void printAllTasks(void);
```
bgamari, line 82: This should be `#if`-guarded by `DEBUG` like its definition site.

PR #267, opened 2022-02-14
fp_prog_ar.m4: take AR var into consideration

10ne1:
In ChromeOS and Gentoo we want the ability to use LLVM ar
instead of GNU ar even though both are installed, thus we
pass (for eg) AR=llvm-ar to configure.

Unfortunately GNU ar always gets picked regardless of the
AR setting because the check does not consider the AR var
when setting fp_prog_ar, hence this fix.

### DIFF
--- a/m4/fp_prog_ar.m4
+++ b/m4/fp_prog_ar.m4
@@ -14,7 +14,8 @@ if test -z "$fp_prog_ar"; then
       fp_prog_ar=$(cygpath -m $fp_prog_ar)
     fi
   else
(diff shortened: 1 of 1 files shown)


10ne1:
@bgamari Hi, gentle ping, any thoughts on this?

10ne1:
I've opened [a MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7725) directly on the GHC gitlab instance, so I'm closing this because there's been no activity.

PR #268, opened 2022-09-09
Fast var set

sgraf812:
Just leaving this here; the effect is mediocre. According to Ticky on the Cabal benchmark, this reduces number of `unsafePersistNode` calls (not sure about the unsafe prefix in hindsight) by 524768 to 524574.

I *think* we could get more leverage by figuring out and focussing on the call sites of `unsafePersistNode` that account for the majority of calls and see whether we can make them amenable to the RULE. But that's a stretch goal; let's focus on your writeup first.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -116,5 +116,5 @@
 	ignore = untracked
 [submodule "libraries/haskell-transients"]
 	path = libraries/haskell-transients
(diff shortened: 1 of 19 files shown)


sgraf812:
Ohh great, wrong repo I opened the MR against 🤡. Closing.

